PR #857, opened 2014-09-19
Moved load_module_package to authenticated ocaps

smschauhan:
Quick request - the load_module_package is mostly used in authenticated scenarios for additional applications. This would be particularly useful for some internal application(hint - go and fish) which require authentication. This is a little critical so it would be very helpful if you could include this in this release.

Thanks,
SMS

### DIFF
--- a/rcloud.support/R/ocaps.R
+++ b/rcloud.support/R/ocaps.R
@@ -57,9 +57,6 @@ unauthenticated.ocaps <- function()
 
       get_users = make.oc(rcloud.get.users),
 
-      # externally used ocaps
-      load_module_package = make.oc(rcloud.load.module.package),
-
       # javascript.R
       setup_js_installer = make.oc(rcloud.setup.js.installer),
 
@@ -133,6 +130,9 @@ authenticated.ocaps <- function()
       # call_fastrweb_notebook...
       call_fastrweb_notebook = make.oc(rcloud.call.FastRWeb.notebook),
 
+      # externally used ocaps
+      load_module_package = make.oc(rcloud.load.module.package),
+      
       # file upload ocaps
       file_upload = list(
         create = make.oc(rcloud.upload.create.file),


gordonwoodhull:
@s-u, I feel like I asked this already but I can't find it anywhere. 

I don't think the rationale given above is sufficient for hiding this OCAP from anonymous users. This is causing att/rcloud.flexdashboard#12 / att/rcloud.flexdashboard#13 - not able to load a flexdashboard as anonymous user.

Can you think of any reason anonymous users should not be able to load packages from the JS side?

s-u:
I think it's ok to expose it but we may want to think about re-writing the code in `rcloud.load.module.package` so that a) it only loads namespaces that are actually modules and b) provide a way that the module can declare whether it's ok to load it in anonymous mode or not.
One quick way to handle both until we have a better solution would be to only allow declared extensions to be loaded in anonymous mode.

s-u:
Actually, now that I'm looking at it I see that `rcloud.flexdashoard` is actually sort of abusing this (i.e., it doesn't actually export any ocaps) so wouldn't the issue go away if you simply added it as a module?

gordonwoodhull:
There is something tangled about this, but yes it does export ocaps

https://github.com/att/rcloud.flexdashboard/blob/master/R/onload.R

I think the idea is that the page needs to load the package in order to produce the ocaps that the page then calls. It's inside-out from how we normally do things.

s-u:
I see, so you don't actually want to load it as a module by default since that OCAP is not really relevant unless you're calling from  `flexdashboard.html`?
I think we're digressing - the fleaxdashboard issue is more related to its design - this issue is about controlling anonymous package loads - which needs to be done otherwise, you can any packages without authentication which is indeed dangerous. So either way we need an explicit way to flag a package as "ok to load anonymously".

gordonwoodhull:
I don't have any other need for loading packages anonymously, other than rcloud.flexdashboard. 

If you can think of a better way for a page to load a package for its OCAPs, I'm all for it.

PR #955, opened 2014-10-20
This Fixes #924 : added padding for highlighting the editable span

### LINKED ISSUE #924: don't let editables disappear
While tagging history version, click on the history name and enter 'backspace' (so as to change the name completely). This will hide the cursor, along with respective history div for that moment until you enter a new name or discard the process.

amol-jore:
Adding the padding of <code>3px</code> to show the editing element


gordonwoodhull:
Eventually applied the same technique only when the contenteditable is not empty, in ed7e2617352560a00db3eedef147d948f35d2405

Issue #1074, opened 2014-11-10
delay welcome message until side panels are in place


gordonwoodhull:
The horizontal sizing of the welcome message is jarring:

![startup-welcome-sizing](https://cloud.githubusercontent.com/assets/1366709/4972175/3f96684e-68a7-11e4-827b-134093726161.gif)

Since it shows before the side panels, it starts out really wide and then gets brought in.  It should not be shown until its width is known.

PR #1140, opened 2014-12-10
Fixed issue #1074 : delay welcome message until it's the right size

sujitbehera27:
Now am holding the welcome message and posting it only when everything else has been loaded.
Taking care of the comment you had posted in #1096.

"I would prefer to handle this as an application problem rather than a UI problem; that is to say, delay the message rather than its container. Nonetheless this is a welcome first attempt and I'll try it out."

Closing PR #1096 and raising new one as its on very old code. Committing there would have shown lots of differences so avoiding merge issues.
@gordonwoodhull Please review and let me know if this is fine.


gordonwoodhull:
Closing this.  Too many hacks built on hacks, and it didn't really help.

Issue #986, opened 2014-10-24
show year in notebook tree dates


gordonwoodhull:
the notebooks may be more than a year old

PR #1227, opened 2015-01-30
Update NEWS.md

martymcwizard:
@gordonwoodhull deleted duplicate line on #986


gordonwoodhull:
Thanks @martymcwizard!  I guess I'm going to kind of merge the two lines, since they are both half-accurate.  The fix affects both notebook dates and version dates.

PR #1245, opened 2015-02-04
Changes to Python side to work with the new RCloud messaging architectur...

sisukapalli:
...e

Main changes in the general RCloud world:
  added a .session parameter to rcloud.support.R tab-completion call.
Changes in the rcloud.python side
  messages are sent as they arrive
  completer implemented (does object level completion, doesn't work with imports, etc.)
  notebook_runner code is cleaned up a bit


gordonwoodhull:
The code looks reasonable, actually somewhat less hacky than before, and I have tested it works.  Merging.

@s-u if you want to do a retroactive code review that would be great.  Got to get this to QA.

PR #1304, opened 2015-02-12
Update conf sample

prateek05:
@gordonwoodhull Put in all the config options I had over my head. The list is not comprehensive. There maybe options I haven't used or don't know off

### DIFF
--- a/conf/rcloud.conf.samp
+++ b/conf/rcloud.conf.samp
@@ -2,86 +2,84 @@
 ##: RCloud 1.3: Configuration Details for RCloud
 ##: RCloud Webserver Details
 ##:----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
seems fine, but why all caps, is it clearer somehow?

prateek05:
Yup

Was working with the spark conf file and everything CAPS makes it more
legible

Moto X@ AT&T

s-u:
errrrr... no :)

s-u:
BTW: you cannot use `##:` because keys have to be unique

prateek05:
:) No to updating ##: again and again. Or the idea of all CAPS ? 
Would it be fine if I rc.cf to ignore all lines starting with '#' ? Can we make '#' as the comment indicator ?

Issue #683, opened 2014-06-06
support binary assets


s-u:
We're half-there with the `.b64` support but it's incomplete. This is to remind me to complete it.

s-u:
Now supported.

Remaining issues on the JS side:

- [ ] the UI doesn't know what to do with binary assets. We could be smart and, e.g., for images show it (`notebook.R` works so they can be turned into links)
- [ ] binary upload into notebook doesn't work because `upload_utils.js` uses `readAsText` (note that `readAsArrayBuffer` does work). There are two ways to fix: a) add "binary" checkbox to use `readAsArrayBuffer` when checked; b) use `notebook_upload` (which supports automatic content detection) instead of `update_notebook` (which is too low-level so it expects caller to make the decision and choose content type accordingly).
- [ ] deleting a binary asset from the UI fails, because `update_notebook` doesn't necessarily know that the asset to be deleted is binary (this could also be addressed by making `update_notebook` more smart)

gordonwoodhull:
For the second item, I think the client could probably read as binary and then convert to a string if it finds the content is text-only. I don't think a checkbox is an option because we support dragging and dropping of multiple assets.

Why does `update_notebook` need to know if an asset is binary? Can't it just send null content?

gordonwoodhull:
Or is it that the server won't know whether the asset is binary if the client just sends null?

s-u:
yes, exactly

for the second item - `upload_notebook` already does that so client shouldn't need to hence the second way forward

gordonwoodhull:
Do you mean `rcloud.upload.asset`? It looks like it detects whether the parameter is raw or string, so the client still needs to detect whether the content is binary and send the correct type.

s-u:
No I mean `notebook_upload` which is an ocap to `rcloud.upload.to.notebook`

gordonwoodhull:
It turns out that it is fine to always read the data as `ArrayBuffer` and update the notebook using `update_notebook`. Then the server decides whether it's binary or not and the client honors that decision.

gordonwoodhull:
@s-u, I put a very ugly hack on the client side to deal with the third point. Sheer laziness. It didn't sound like the hack on the server side would be much less ugly, and probably less efficient. But if you feel like cleaning up the interface, the guilty commit is the change to asset_model.js in 1f6c39aac64e4deaa3bbc553e54f1d93fc1bdfbf

gordonwoodhull:
[...]

Issue #1059, opened 2014-11-05
Error in uploading certain files as Assets


sneha-bharti:
Steps to be followed to test -
1) Upload a file, namely "d3.js"
2) The file gets uploaded temporarily but the file-upload div throws an error "(unexpected) undefined" and the session div displays "Socket was closed"
3) Reloading the page, opens a fresh edit.html page of the notebook but without the file "d3.js"
This error is not shown while uploading other files as shown in the screenshot. Probably the size of the file is causing the problem.
![assetupload 2](https://cloud.githubusercontent.com/assets/5227441/4913975/439bce92-64bb-11e4-98d2-047cb5e1d090.jpg)

gordonwoodhull:
Was not able to reproduce on 3 OS x 2 browsers.  However, I was able to reproduce #984 file upload using d3.js as the test file.

Will keep this open but put it on Unknown until someone reports an operating system and browser where they are seeing this.

PR #1323, opened 2015-02-14
Feature/binary assets

s-u:
Adds programmatic support for binary assets

Addresses in part #683

What is missing is support for binary file upload, but that will require UI support to declare binary uploads (checkbox?) since they need different handling. Also binary assets are likely large and direct upload breaks for anything that is not small.


gordonwoodhull:
I have received conflicting reports, but in my testing, uploading assets of any size does not break (#1059 is the open issue, was not able to repro).

Can we just detect binary with a simple [/[\x00-\x1F]/](http://stackoverflow.com/questions/1677644/detect-non-printable-characters-in-javascript)?

gordonwoodhull:
What part of #683 does this not address?

rcloud.support/src/json.c
```diff
@@ -0,0 +1,323 @@
...
+    /* unreachable */
+    Rf_error("invalid element type");
+    return R_NilValue;
+}
+
+SEXP parseJSON(SEXP sWhat) {
+    const char *c = CHAR(STRING_ELT(sWhat, 0));
+    SEXP res = parse(&c, c + strlen(c));
+    return (TYPEOF(res) == CHARSXP) ? ScalarString(res) : res;
+}
```
gordonwoodhull: No!  You didn't!

gordonwoodhull:
The problem that I have on the client side, is that we have kind of a strict MVC scheme where I always update the notebook through update_notebook.  I have temporarily tried to use `rcloud.upload.asset` instead - no luck so far, but even if I get it working, I will have a problem of models and views and controllers out of sync with each other. I also don't know if our buffer/asset model can handle binary.

I would rather stop using JSON across the wire, as we have discussed, but that's a bigger change.

Let's talk later today.

rcloud.support/R/rcloud.support.R
```diff
@@ -270,13 +274,45 @@ rcloud.unauthenticated.notebook.by.name <- function(name, user=.session$username
...
+      ## in this case we have to make sure we delete any TXT version
+      files <- list(x=NULL)
+      names(files) <- gsub("\\.b64$", "", name)
+      .binary.to.b64(file)
+  } else if (is.null(attr(file, "metadata")) && ## the contents may not have metadata for text storage
+             checkUTF8(file, quiet=TRUE, min.char=7L)) { ## and has to be valid UTF-8, otherwise we use b64
+      rawToChar(file)
+  } else { # auto-convert to .b64
+      ## in this case we have to make sure we delete any TXT version
+      files <- list(x=NULL)
```
gordonwoodhull: I think it fails because github won't let you delete a file that's not there. At least I'm getting `Validation Failed`

PR #1332, opened 2015-02-15
Fixes #1327 Show first 5 values

### LINKED ISSUE #1327: Testing on RCloud local: Variable values are missing in workspace div
For the any variable with length equal or greater than two, only the variable name, its data type and length are displayed in workspace div. The values are missing for such variables 

workspace div

the behavior is as expected for the variable with single value

prateek05:
@gordonwoodhull Hope this helps
Values
```c	character[1:3]	string1,string2,string3```
```d	character[1:6]	string1,string2,string3,String4,String5```
```ddd	integer[100 x 1]	1,2,3,4,5```

### DIFF
--- a/rcloud.packages/rcloud.enviewer/R/rcloud.enviewer.R
+++ b/rcloud.packages/rcloud.enviewer/R/rcloud.enviewer.R
@@ -12,9 +12,6 @@ rcloud.enviewer.view.dataframe <- function(expr)
 rcloud.enviewer.display.dataframe <- function(x)
   list(command="view", object=x)
 
-rcloud.enviewer.display.double <- function(x)
-  signif(x,3)
-
 rcloud.enviewer.display.string <- function(x)
   paste0("'", as.character(x), "'")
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by gordonwoodhull** [comment]
I'm merging this, but there is still plenty more we can do.

For example., for the matrix case `m <- matrix(runif(4*4), nrow=4)`, we get the entire data, because `head(m, 5)` takes 5 rows.

R displays the very elegant

```
          [,1]       [,2]      [,3]      [,4]
[1,] 0.7976235 0.07294365 0.1288330 0.1354370
[2,] 0.7669677 0.98953275 0.4887348 0.6021054
[3,] 0.3463109 0.21583140 0.7708296 0.5930679
[4,] 0.7512210 0.93857457 0.1628632 0.6383253
```

We'd probably want to show the first row in this case and then display the whole matrix in the Viewer, with a link from the type **double[4 x 4]**

But that's all later.  This is a fine start.

prateek05:
Your are right. We can handle tabular data more better. Will aim for the next point release

PR #1350, opened 2015-02-18
Update INSTALL.md

shyamram:
correct a very minor typo, mainly to freshen the date of the page


gordonwoodhull:
Um, I don't see how a week old is too old? 

This document really is up to date and is the one that people should use.  

But okay, merging...

PR #1359, opened 2015-02-20
Fix typo gitgist

prateek05:
@s-u  Hi Simon. Please review contains a fix of misspelled gigistcontext in the gitgist package


s-u:
Thanks, I have committed a subset of this.

prateek05:
@s-u  the require(github) in login_successful.R is not really required.
https://github.com/prateek05/rcloud/commit/4441c46613cf22308cba5e7fa6bf9d47bae1b84c

s-u:
I bet that's not the only thing not required there ;)

prateek05:
It's slightly annoying to have loading github message on the invoking shell session whenever someone logs in. It traces back to here.

s-u:
But this is only if you don't use github, right? It's still needed if you do (but loaded via gist -> githubgist ...).

s-u:
I'm just a bit wary of unnecessary changes at this point ...

prateek05:
Precisely its loaded via gist->githubgist and isn;t really required even when using github as the backend. 
I am assuming its left over from the era previous to introduction of different backends.
Also agree with second comment. Perhaps after the patch release a rewrite of the code needs to take place.

PR #1361, opened 2015-02-21
Filter rcloud extensions

gordonwoodhull:
introduce notion of "session type"
    
perhaps a poor name, but the idea is that when we initialize a session, we say what kind of session we need, and only the appropriate extensions are loaded

extensions are tagged in their `DESCRIPTION`, key `RCloud-Extension`

@s-u, i know my R is probably crap, but does this look decent otherwise?  

"session type" is perhaps vague: it is actually a set of tags, and if any match, the extension is loaded.

so edit mode has tags `['edit', 'gui']`, view mode has `['gui']` only. mini and shiny have none.


gordonwoodhull:
I closed this because adding a parameter to `*session.init` breaks the interface and causes a lot of problems.  Maybe moving it to a new ocap `rcloud.load.extensions` could help.

PR #1379, opened 2015-03-04
fiddleR

bartbutler:


gordonwoodhull:
@shaneporter, here's an interesting one for when you're back, or if you want an interesting design problem (architecture and UI) while you're on vacation.

A really bright programmer from another division of our company came out to the Hackathon at the company Software Symposium a year ago, and built the tool that he wished he had when he was working with RCloud.

He had built a very complex dashboard using a version of mini.html and [rcloud.dcplot](https://github.com/att/rcloud.dcplot) and got engrossed with what it's like trying to build an application from the HTML and JavaScript side, with less emphasis on R code, or with the R code provided by others. 

Much as you've been doing with RCAP, but without the time to build a framework.

So the UI he designed in this PR is for when you are only working on the layout and html/css/js of a notebook. The command-prompt / editable-markdown interface of the notebook itself is gone: instead the live dashboard output is in the notebook area in an iframe, and you work in the assets pane with your modified mini.html in an asset, plus JS and CSS and R assets; and you refresh the output frame as you fix stuff.

Given this raw hack of an implementation, put together over a night or two at the Hackathon, and your experience with the code of RCloud core, how would you merge this or implement this differently? How would you generalize this, and does this suggest that RCloud needs other extension points?

gordonwoodhull:
BTW, yes, as the title implies, this brings rcloud closer to JSFiddle (or [Building Bl.ocks](http://blockbuilder.org)).

gordonwoodhull:
@shaneporter, this one might have slipped past your radar because it's a pull request I want you to review, rather than an issue. Please take a look when you have time.

I will also find more issues and determine where we are for the release, but that might not be tonight because I've worn myself out at a conference the past few days.

shaneporter:
It's beeping on my radar; will take a look in the next 24 hours.

shaneporter:
It's been a while since I commented on this. I did take a look at it, but couldn't get things working because of an apparent R issue (it couldn't find a resource.)

I'm going to attack it from a different side and see how I get on.

gordonwoodhull:
Hoping we can get to this in the next point release.

shaneporter:
When is 1.6.1 due to be released?

gordonwoodhull:
There is no particular schedule but I think September is a reasonable target. It also may be that what's now marked 1.6.1 will become 1.6.2 if there are things that need to be patched in 1.6. Right now I'm using it as a bucket for "what's next".

shaneporter:
I know I mentioned this on a call, but stating here that I got this working.

PR #1450, opened 2015-05-13
fix #1401 left margin of inactive markdown is crazy huge

### LINKED ISSUE #1401: left margin of inactive markdown is crazy huge
When notebooks div and assets div are minimized the, the code lines in markdown cell don't get shifted as seen in the screen shot below
 
markdown

anatoliyg:


gordonwoodhull:
Thanks!

PR #1451, opened 2015-05-13
fix #1387 - adding HTML highlighting of assets

### LINKED ISSUE #1387: HTML highlighting
in asset pane

anatoliyg:


gordonwoodhull:
For the "define is not defined" error, it looks like we need to use the built version of ace, not the sources. 

http://stackoverflow.com/questions/10883190/ace-editor-define-is-not-defined

I think this may be the step I forgot about. I have been building ace and then copying files from one of the subdirectories. I'd have to check my other computer to remember which one, but the question points to a built version which should work for this single file.

anatoliyg:
Ok , got it. Thanks , Gordon

gordonwoodhull:
Thanks, went ahead and built ace so I could remember how to do it, copied in the html worker (lots bigger with all the dependencies), and merged.

anatoliyg:
Thanks. I had the right version of the worker locally, forgot to check it in

PR #1484, opened 2015-06-10
notebook protection v1

anatoliyg:


htdocs/css/bootstrap-rcloud-tweaks.css
```diff
@@ -306,3 +306,435 @@ td {
...
+    padding-right: 10px;
+}
+
+.selectize-input{
+    min-height: 100px!important;
+
+}
+
+
+/**
```
gordonwoodhull: please include selectize.default.css as a separate file in the css/ directory

htdocs/editor_tab.js
```diff
@@ -413,6 +413,14 @@ var editor = function () {
             rcloud.stars.get_my_starred_notebooks()
         ])
             .spread(function(user_notebook_set, my_stars_array) {
+
+                rcloud.get_users().
+                then(function(data){
+                    window.allTheUsers = data;
+                });
+
```
gordonwoodhull: let's figure out how to only call `get_users` once, since it's expensive. Means refactoring the `.all()` above.

htdocs/js/notebook/notebook_controller.js
```diff
@@ -89,6 +89,12 @@ Notebook.create_controller = function(model)
             }
             model.user(notebook.user.login);
             model.update_files(notebook.files);
+
+            //
+            RCloud.UI.notebook_protection.userLogin = notebook.user.login;
+            RCloud.UI.notebook_protection.userId = notebook.user.id;
```
gordonwoodhull: you probably want to use editor.username() instead of this, which is the notebook user. it looks like (will test in a minute) if you load someone else's notebook, then the protection dialogs will be populated from that user instead of the logged-in user. also `userId` is unused.

htdocs/js/ui/init.js
```diff
@@ -1,4 +1,6 @@
 RCloud.UI.init = function() {
+
+
     $("#fork-notebook").click(function() {
```
gordonwoodhull: just a nit, but please look out for added newlines. it just adds noise to the changelog.

htdocs/js/ui/notebook_protection_controller.js
```diff
@@ -0,0 +1,620 @@
...
+   .controller('NotebookProtectionController', ['$scope' , 'GroupsService', function ($scope, GroupsService){
+
+
+   		//Notebook tab vars
+   		///////////////////////////////
+
+	   	//user 
+	   	$scope.userLogin = '';
+		$scope.userId = '';
+
```
gordonwoodhull: another nit: please use spaces not tabs. tabs display very differently on different computers.

htdocs/js/ui/notebook_protection_controller.js
```diff
@@ -0,0 +1,620 @@
...
+				var conf = confirm("Are you sure you want to make your notebook public?");
+				if(conf){
+					console.log('notebook id is '+$scope.notebookGistName);
+					GroupsService.setNotebookGroup($scope.notebookGistName, null )
+					.then(function(data){
+						console.log('data is '+data);
+					})
+					.catch(function(e){
+						
+						$scope.cancel();
```
gordonwoodhull: this looks mismatched - should it be in the success section?

htdocs/js/ui/notebook_protection_controller.js
```diff
@@ -0,0 +1,620 @@
...
+				if(r === true){
+
+					//var groupId;
+	                var prevGroupId = $scope.selectedGroup2.id;
+	                GroupsService.changeGroupName($scope.selectedGroup2.id ,pr)
+	                .then(function(data){
+	                    //groupId = data;
+
+	                    var theIndex;
+						for(var i = 0; i < $scope.allUsersGroups.length; i ++){
```
gordonwoodhull: using something like `_.find` would reduce the amount of code. I'm seeing this loop repeated quite a few times.

htdocs/js/ui/notebook_protection_controller.js
```diff
@@ -0,0 +1,620 @@
...
+
+	    //Admins and Memebers areas
+	    ///////////////////////////
+
+
+	    $scope.startWatchingGroups = function(){
+
+	    	var count1 = 0;
+	    	var count2 = 0;
+
```
gordonwoodhull: are these counts just obsolete sanity checks?

anatoliyg: For some reason the callback for watch function is firing right away, even thou they are only supposed to fire when the model changes. So the counts are used to not react to the first callback for watch. Once I figure out why this is happening, I can remove the count variables. Probably has something to do with using $scope.$evalAsync

gordonwoodhull:
Gave it a quick read. Looks good, aside from a couple style-guidelines kind of things I noted in comments.

I'm going to merge this, and we can continue to revise it.

anatoliyg:
Copy

gordonwoodhull:
Okay, I have some minor comments, but there is only one major one. 

I opened issue #1485 - I think it will be difficult to test if users can't assign notebooks to a group unless they're an admin.

PR #1488, opened 2015-06-11
Fixed some UI issues / Admin priveleges

anatoliyg:
selects filtered by admin privileges
going public from private
first time UI and cases when you have no groups or delete all your groups.
group selected when created
small fixes


gordonwoodhull:
Thanks! I deleted the pathetic attempts at click-away and added a close button to the popover, so it's useable, and merged. We've got a ways to go but at least it seems to generally work.

PR #1500, opened 2015-06-12
fixed issue where new group members were not pupulated

anatoliyg:


htdocs/css/bootstrap.css
```diff
@@ -3867,8 +3867,8 @@ textarea.input-group-sm > .input-group-btn > .btn {
 }
 
 .nav-tabs > li.active > a,
-.nav-tabs > li.active > a:hover,
-.nav-tabs > li.active > a:focus {
+.nav-tabs > li.active > a,
+.nav-tabs > li.active > a {
```
gordonwoodhull: we avoid changing bootstrap.css, and any other external dependencies, directly - which is the original reason we have bootstrap-rcloud-tweaks.css. I'm moving this rule to bootstrap-rcloud-tweaks.

gordonwoodhull: although i see this was not the first change..! reverting those too.

htdocs/js/ui/notebook_protection.js
```diff
@@ -1,88 +1,113 @@
...
-    this.passedData = null;
-    this.appInited = false;
-    this.theApp = null;
-
-
     //set from outside
+    this.defaultCryptogroup = null;
+    this.defaultNotebook = null;
     this.userId;
     this.userLogin;
```
gordonwoodhull: these two lines don't do anything. changing them to initialize to null.

anatoliyg: for sure, all of those types of declarations should be initted with either null, '', [] or {}.

Issue #1143, opened 2014-12-11
support identify()


gordonwoodhull:
now that we are going to have live graphics devices for each graphic, it should be possible to support `identify` to label points as they are clicked on

gordonwoodhull:
oob message: device id, page id. returns x,y or NULL

gordonwoodhull:
mouse cursor crosshairs

s-u:
Implemented

PR #1505, opened 2015-06-15
first stab at a locator()

gordonwoodhull:
for #1143

Hey @s-u, here are most of the javascript bits for a locator. It takes a device id and page explicitly because I don't remember how to get the current device. It pushes the pixel coordinates to an array and then sends the array when ESC is pressed.

However, something is wrong with the continuation (or something), because I don't see the value returned, whether I pass it first or second to `k()` - the usual RCloud integration shenanigans, I'm sure.

Obviously I didn't worry about how to integrate it or translate coordinates or any of that good stuff, because I figure you know all that. I've actually been testing it in the debugger, watching the image created to get the device id, then running `rcloud.locator` with the device id. 

Theoretically there would be no problem running this on any image in the notebook, not just the current one...


gordonwoodhull:
LMK what other interface you want on it. I actually think the blue keyboard-focus highlight works pretty well. Oh, I'll add the crosshair cursor right now.

gordonwoodhull:
Alright, there are intricacies about getting the right coordinates:

http://stackoverflow.com/questions/2159044/getting-the-x-y-coordinates-of-a-mouse-click-on-an-image-with-jquery

I think I chose the best answer there (instead of the most popular one). Of course, I'm just returning coordinates relative to the image, not the plot.

Also, crosshair! For the duration of the locate!

Done now, zzz!

s-u:
Awesome! Taking over :)

s-u:
Ok, I think we're ready to merge now ...

gordonwoodhull:
Go for it and tell QA how to test it!

PR #1603, opened 2015-07-23
Automated Test Scripts

sanketd11:
Added the automation test scripts.


gordonwoodhull:
Thanks @sanketd11 and team! I'll try to review and merge this for 1.4.3

gordonwoodhull:
@anatoliyg briefly reviewed the code, and ran the tests on OSX after porting the dependency installation script. We like it. 

I'll review the code again and we can add the other install script.

gordonwoodhull:
The one thing I still don't like is the use of XPath. I find those expressions impenetrable and I fear they will make the tests hard to maintain.

Apparently Casper does support CSS3 selectors: http://casperjs.readthedocs.org/en/latest/selectors.html#index-1

Is there a reason they wouldn't work, or was it just what was easy to glean from the browser debugger versus what would take reading the HTML source to figure out?

gordonwoodhull:
Also, @anatoliyg reports that many of the tests failed for him, so we either have a problem configuring/installing the dependencies, or the tests aren't completely portable.

prateek05:
@gordonwoodhull thats because the wget is of the binary for linux and not really being compiled from source phantomjs-1.9.8-linux-x86_64.tar.bz2

gordonwoodhull:
@anatoliyg and I did a fresh install and test today on OSX, and had no problems this time. We ran about 4-5 tests. Must have been a bad install of PhantomJS last time, or something. 

We did not use the installation script, just brew and npm. Since that is going to be so very platform-specific, it might be of limited usefulness.

The only weirdness we were seeing was that we had to fill in the username&password manually but maybe we were not patient enough? Once it got past that, automation worked great!

gordonwoodhull:
I think it might be better just to specify the dependencies and not include an installation script, since there will be differences on every platform and every machine. I didn't merge this for 1.4.3 because you haven't addressed @prateek05's comments, and we do need instructions for OSX, where most of us do development.

It might be better to merge this to develop, get a Jenkins server somewhere (and see how we do with automatic testing of commits), and officially release this in 1.5. 

This is great work.

gordonwoodhull:
Team, could you please update the PR by removing the installation script and describing in English the installation of the dependencies in `Readme.md`. Similar to how we describe how to install RCloud in [INSTALL.md](https://github.com/att/rcloud/blob/develop/doc/INSTALL.md)

I think it's better not to ask people to run a script as sudo, and instead let them use their judgment how to best install the prerequisites. Copying and pasting some commands from Markdown just feels safer to me.

Then I will merge to develop and we'll add similar instructions for OSX - some of these needed to be installed using Homebrew and others with npm, I can't remember which were which.

gordonwoodhull:
Sorry this didn't make it into 1.5 - I didn't see the updates in time. We'll try to merge this for 1.5.1.

gordonwoodhull:
The instructions still say to use `sudo` to run tests. That seems weird. I'm going to try without that, once these dependencies install.....

gordonwoodhull:
Hmm, I just get nothing happening, whether I use `sudo` or not. 

Installed everything through homebrew. 

Will have to keep testing to see how to refine the OSX instructions better.

gordonwoodhull:
I'm going to take @s-u's suggestion and merge this to develop instead, for 1.6. Then we can get automatic testing working for the develop branch, where it really helps, and it will be hopefully be really solid for that release. 

Sorry for the delay but I think this is the most careful way to go about this.

gordonwoodhull:
Although I haven't gotten this to work personally, I know @anatoliyg has, and I think this is safe to merge to develop. We'll make sure the OSX instructions are clear for the 1.6 release

PR #1609, opened 2015-07-30
Filter rcloud extensions, take 2

gordonwoodhull:
This replaces #1361.

Currently we are only deactivating rcloud extensions for notebook calls. We should allow each session to specify, when it is initialized, which types of extensions should be loaded.

I have resurrected an old branch for this, which was not stable enough for 1.3. The idea is to have a set of flags in the DESCRIPTION file of each RCloud extension, which say what kinds of functionality the extensions provides. Then, after doing the usual per-instance and per-user filtering, these flags are intersected with the flags specified for session.init.

This is ***not ready for merge***, since it introduces an unnecessary extra flag which was changing function signatures and wreaking havoc everywhere. As suggested by @s-u, we already have a flag for this, but that one needs to be upgraded to a vector of flags. So far I've just merged develop into this branch, so that we have an active PR for 1.4.3.


gordonwoodhull:
What @s-u and I discussed... a long time ago... was to use the existing `mode`, which is currently a 'scalar' string
 * change it into a vector of flag-strings
 * combine that functionality with the extra flags that are being sent along with the very breakful `session_type` in this PR
Then the same flags will be used to determine whether to do compute separation and to decide which packages to load.

This means e.g. changing 
```{r}
    if (mode %in% cs.modes) { ## use fork only in modes that require it
```
https://github.com/att/rcloud/blob/develop/rcloud.support/R/ocaps.R#L88

to a union operation, etc etc. And removing most of the virulent interface changes in the current PR (yay).

gordonwoodhull:
I'll clean this up soon and pass back to @s-u for review.

gordonwoodhull:
This seemed potentially too disruptive for the point release - could have strange effects for third-party extension developers, and we also need to make sure they know about the tags they need for the DESCRIPTION file. For now, we've disabled extensions also for mini.html (see commit above).

Hope to fix this up and merge to develop soon for 1.5.

PR #1669, opened 2015-09-21
rcloud.output.context

gordonwoodhull:
Creating this in a feature branch since we don't have a release branch for 1.5, and this is for 1.6.

Output contexts worked on a first try. Just needed to attach the OOBs to mini.js, which seems reasonable.

Three things which are not reasonable:

 - [ ] get @s-u to design the API
 - [x] plots always get appended after html output, is this the usual behavior? 
 - [ ] adding jquery-ui.css to mini.html somewhat questionable
 - [x] we really need `rcloud.flush.plot`, because right now you need to create an extra plot in order to force the last one to display.
 - [x] the default bootstrap.css (which we load in mini.html why?) shows code in red, so `rcloud.out` output looks like an error.
 - [x] investigate whether errors get displayed properly; in a brief test I didn't see them.
 - [x] although there shouldn't be any noticeable resource usage, we should have a close/release command.


gordonwoodhull:
Simple example code for a mini.html notebook:
```{r}
require(rcloud.web)

rcw.result(run=function(...) {
    context.id <- rcloud.output.context('#test')
    Rserve.context(context.id)
    rcloud.html.out('<p>This should look like HTML</p>')
    rcloud.out('This should look like code')
    plot(1:9)
    plot(1:9)
    },
    body="<div id=test><p>We should append stuff here.</p></div>")
```

gordonwoodhull:
Example of two simultaneous output contexts:

```{r}
require(rcloud.web)

rcw.result(run=function(...) {
    context.id1 <- rcloud.output.context('#context1')
    context.id2 <- rcloud.output.context('#context2')
    Rserve.context(context.id1)
    rcloud.html.out('<p>We can write to two contexts</p>')
    plot(1:9)
    Rserve.context(context.id2)
    rcloud.html.out('<p>Here is the second one</p>')
    plot(9:1)
    Rserve.context(context.id1)
    rcloud.html.out('<p>Continuing on first</p>')
    plot(1:17)
    # we need flush
    plot(1:1)
    },
    body="<table><tr><td><div id=context1><p>Output Context 1</p></div></td><td style='vertical-align: top;'><div id=context2><p>Output Context 2</p></td></tr></table>")
```
Arguably, it might be more R-like to automatically assign the context when creating it. I'll defer to @s-u on what the API should look like.

gordonwoodhull:
A fourth unreasonable thing: the default bootstrap.css (which we load in mini.html why?) shows code in red, so `rcloud.out` output looks like an error.

gordonwoodhull:
Also investigate whether errors get displayed properly; in a brief test I didn't see them.

gordonwoodhull:
Also, although there shouldn't be any noticeable resource usage, we should have a close/release command.

htdocs/js/session.js
```diff
@@ -252,7 +252,9 @@ RCloud.session = {
     }, init: function(allow_anonymous) {
         this.first_session_ = true;
         return rclient_promise(allow_anonymous);
-    }
+    },
+    on_data: on_data,
+    on_oob_message: on_message
```
gordonwoodhull: this looks wrong

gordonwoodhull:
Flushing seems to allow control over how plots & html combine.

gordonwoodhull:
Adding `bootstrap-rcloud-tweaks.css` to mini.html to fix output formatting.

gordonwoodhull:
Errors kill execution and get printed to the debugger console. I think this is to be expected given the way mini.html notebooks are executed; however, it's different from how cells work.

gordonwoodhull:
- [ ] Save image as.. menu is missing

gordonwoodhull:
I'm merging this and will open an issue for the Save Image command.

Some of the extra dependencies here are maybe still questionable, and @s-u will probably change the API, but this is what develop is for. Please open issues with any questions or concerns!

PR #1735, opened 2015-11-03
WIP - load all notebook info at once

gordonwoodhull:
IIRC, one of the bottlenecks we were finding is uploading the list of notebooks and users from the client to the server, which is something we shouldn't have to do.

This optimization, which sends the big list only from server to client, seems to speed up rcloud-2 and rcloud-5 by about 4-5 seconds. It may help dramatically in VPN situations.

Not the prettiest fix, still needs work, and the function name `rcloud.config.all.users.all.notebooks.infos.stars` is purposely horrendous... just an experiment to improve on.


gordonwoodhull:
I think this is fine. Renaming the API to something reasonable and merging.

Doesn't solve the bigger notebook performance problems but should help folks on VPNs.

PR #1782, opened 2016-01-20
first draft of undo/redo feature for notebooks using standard keyboard shortcuts

shaneporter:
This is a set of changes to support @gordonwoodhull's suggestion to implement undo and redo commands with standard key commands. CTRL/⌘ + z and CTRL/⌘ + y.

The commands are shortcuts that step through the notebook's history, but they do not commit or create a new version. It will still be the user's responsibility to click the 'revert' button.

Note that when the focus is with Ace editor, the editor's standard undo/redo shortcuts will be operational. The notebook's undo/redo functionality implemented with these changes will only work with a 'global' focus.


gordonwoodhull:
Thanks @shaneporter!

The minor refactor of the `keydown` handler is great, especially since our users want us to add a lot more keyboard shortcuts this year.

However, can you find a way to execute the action that is semantic, rather than simulating a click? Simulating clicks is somewhat brittle and leads to code that is harder to maintain.

I can see why you did it this way, because the "right way" involves going a bit deeper in the code and exposing a history API. I realize that right now there is no API and we are just building the tree directly from the data, which is a poor design. 

There's no need to build out the whole API, but just add what you need. Rather than implementing `get_selected_node`, use `current_.notebook` and `current_.version`. Rather than simulating a click, look at what `tree_click` actually does, and implement enough of an API to get the information from the `$tree_` to do that.

The current solution doesn't work until the history list has been opened. It should work even before the user has opened the tree. This means you need to `show_history` before undo. Further, `show_history` is currently not returning the promise that it creates; you'll need to change that so that you can wait for the data to be retrieved and the history shown.

There are also a couple of UI quirks:

 * the usual redo command on Mac is cmd-shift-Z, not cmd-Y. You can look at the end of `htdocs/js/ui/find_replace.js` for how we do a platform-specific key mapping for "replace", which is ctrl-H on Windows but cmd-opt-F on Mac.
 * you should `e.preventDefault()` (and usually `e.stopPropagation()` for mouse events) whenever handling an event, to make sure that nothing else handles the same event. On Chrome/Mac, cmd-Y redo is working but also opening Chrome's own history window in another tab.

shaneporter:
Hi @gordonwoodhull,

thanks for your comments - I've made some changes, but am yet to tackle your main suggestion - the history API.

Is it best for the code changes for this to be in editor_tab.js? If so, do you suggest changing the tree_click code so that it references the (new) history API's methods based on the notebook and version? The redo and undo shortcuts could then call methods on this history API to update the current state (and tree).

Wanted to get some clarification before I dive in.

Thanks!

gordonwoodhull:
I'm fine with almost any code organization that reuses functionality through functions rather than generating events on other UI elements. 

I'm not sure what you mean by `tree_click` using the history API. Right now it doesn't really use the history structure at all - it just reads the notebook and version from the node which was clicked. 

I think all that's needed for a history API right now is something like `next_notebook_version` and `last_notebook_version`. Then `step_history_undo` and `step_history_redo` can use these functions and `current_.notebook` and `current_.version` to determine what to navigate to, and then call `open_notebook`.

And yes, I'm just imagining an internal API within editor_tab.js for now - as I'm sure you've noticed working on RCAP, there are always lots more things which could be exposed in RCloud... but for now editor_tab is kind of a monolith and we don't really have time to break it up and expose functions as we should.

gordonwoodhull:
This looks great! I hadn't thought of using `histories_`, that is a nice solution.

FYI, GitHub doesn't alert me when there has been an update to a PR, so it's best to leave a comment when updating, to ping me. Looks like this is not yet completely hooked up (`output_history_debug`) so maybe that was on purpose...

shaneporter:
Thanks for the info; I've removed `ouput_history_debug`.

htdocs/editor_tab.js
```diff
@@ -1402,7 +1445,7 @@ var editor = function () {
                 }
                 if(opts.toggle) whither = 'hide';
             }
-            update_history_nodes(node, whither, null)
+            return update_history_nodes(node, whither, null)
```
gordonwoodhull: This turned out not to be necessary with the refactor, since now we don't have to show the menu in order to act on it. It's still a good idea to return promises wherever they are generated, even if they're not currently used.

htdocs/js/rcloud_bundle.js
```diff
@@ -6420,6 +6424,16 @@ RCloud.UI.init = function() {
             range.setStartAfter($('.response')[0]);
             selection.addRange(range);
         }
+        // undo
+        if(isCmdOrCtrlAndKeyCode(90)) {
+            e.preventDefault();
+            editor.step_history_undo();
+        }
+        // redo
+        if((!ui_utils.is_a_mac() && isCmdOrCtrlAndKeyCode(89)) || (ui_utils.is_a_mac() && e.keyCode == 89 && e.metaKey && e.altKey)) {
```
gordonwoodhull: this is cmd-opt-y in the mac case, when it should be cmd-shift-z. i'm going to fix this and go ahead with the merge, since the rest looks good!

htdocs/editor_tab.js
```diff
@@ -1412,9 +1455,24 @@ var editor = function () {
                         $(".history i",$(node.element)).addClass("button-disabled");
                     }
                     $tree_.tree('openNode', node);
+
+                    return Promise.resolve();
```
gordonwoodhull: This is returning a promise of undefined instead of just returning undefined, so it doesn't really do anything.

htdocs/editor_tab.js
```diff
@@ -1057,6 +1067,39 @@ var editor = function () {
...
+            var current_index = current_.version === null ? 0 : this.get_current_notebook_history_index();
+
+            if(current_index === this.get_current_notebook_histories().length - 1) { 
+                return undefined;   // already at first
+            } else {
+                return this.get_history_by_index(current_index + 1).version;
+            }
+        },
+        get_next : function() {
+            var current_index = this.get_current_notebook_history_index();
```
gordonwoodhull: This fails in the on-current-version case. I'll open a separate issue for that.

gordonwoodhull:
#1786 is the redo bug

Issue #1086, opened 2014-11-14
print all the code cells when looking at an un-executed notebook


holyguster:
A user asked for this.  Some folks like to print out code.

Right now, you can print what is showing on the screen, but code in code cells that is not showing on the screen will not be included (say, there is a scroll bar, the code for which you need to scroll down to see will not print).  Also, if the notebook is longer than my browser's current window view of the notebook, only what is showing in the browser will be printed.

The scroll bars seem to be the obstacle, it's almost like the print function should expand all cells/frames such that the scroll bars are not needed.  Not sure how that works....

gordonwoodhull:
Looks like the easiest way to support this is to add a special style sheet:

https://developer.mozilla.org/en-US/docs/Printing

gordonwoodhull:
Related: export to PDF #1320, although it might be annoying to have your code print with a grey background when it's just code and no results.

gordonwoodhull:
Sorry, this did not happen for 1.3 but I'll make sure to get to it for 1.4

gordonwoodhull:
As a workaround, you can copy the entire code to the clip board using the mouse and print it out in another app. I could probably add ctrl/cmd-A to make this easier.

gordonwoodhull:
#1321 select all is in place.  the workaround is: select a little bit of the notebook, then hit ctrl/cmd-A to select all, then copy and paste in another app to print.

gordonwoodhull:
A user reports:
> I don’t have access to the source code, I removed all css via the chrome developer tools. Then add the following:
```
body, html, #wrapper {
  width: 100%;
}
```
 
> I decided it was possible after reading, http://stackoverflow.com/questions/10845621/window-print-is-not-printing-the-whole-page.

gordonwoodhull:
Implemented in #1783

gordonwoodhull:
Feedback on the formatting and page break behavior is welcome.

PR #1783, opened 2016-01-25
Feature/print notebook

shaneporter:
I've created an initial print stylesheet, focusing on the cell area (middle column).


gordonwoodhull:
Awesome! This is really close. Couple things I noticed:

 - The welcome message is part of the session, but I don't think folks will want it as part of the printed output. I'm not sure about this, but I'd be inclined to hide it and see if people complain that it's not there, rather than the other way around. It's something they didn't create or ask for.
 - It seems to be over-eager on the page breaks, like it's prioritizing keeping cells or cell output together, rather than minimizing the page count. I think the expected behavior would be to break in the middle of cells, wherever the page runs to.

shaneporter:
Hi @gordonwoodhull - please have a look at my changes, hopefully they address your comments. 
Thanks.

gordonwoodhull:
This addresses #1086

gordonwoodhull:
Looks great! We may still have to do some fine tuning of styles (should it look more like Markdown, with grey code?) and pagination (should cells be kept in one piece in some conditions?) but we can put it out and get feedback on that stuff.

I'm also applying the stylesheet to view.html - I always forget, but generally view.html should be tested whenever edit.html is changed, because they share a lot of code.

PR #1788, opened 2016-01-27
fix for #1786 - get history index updated

### LINKED ISSUE #1786: "redo" on current version throws error and goes to last version
@shaneporter, a little glitch here - when you hit the redo key on the current version, it looks like it's searching for `null` in the history, not finding it, and then not handling the `-1` correctly:

```
Uncaught TypeError: Cannot read property 'version' of undefined
```
https://github.com/att/rcloud/blob/develop/htdocs/editor_tab.js#L1098

Then it goes to the penultimate version as if "undo" were requested.

shaneporter:
`get_current_notebook_history_index` wasn't accounting for a null value for `current_.version`; I have changed the code so that it now does so, and this has fixed the 'latest version redo' bug.

this fixes #1786

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -1072,9 +1072,11 @@ var editor = function () {
             return histories_[current_.notebook];
         },
         get_current_notebook_history_index : function() {
-            return find_index(this.get_current_notebook_histories(), function(h) {
-                return h.version === current_.version;
-            });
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Sorry, this fixes the error, but redo still goes to the penultimate version. 

I haven't looked into why this is happening, but there must be another bug here.

Issue #658, opened 2014-05-14
Crop / cull cells


gordonwoodhull:
Often when trying stuff out you end up with a bunch of junk cells.

There should be a command to select which cells you want to keep, and delete the rest.  Since we support "sortable" dragging and dropping of cells, why not treat cells as items more thoroughly:

 - when you click on a cell title/gutter it becomes "selected" with a border
 - you can cmd/ctrl-click to add to the selection
 - shift-click to select a range
 - (there are confusing ctrl-shift-click standards we could consider for the obsessive users)
 - <delete> deletes those cells
 - look at what Crop is usually mapped to (as in #300, naming is hard and we should look for standards). Crop is the inverse of delete, deletes every cell except those selected.

In order for the key commands to work, we'll have to make sure that all ace editors get deactivated when these sorts of commands happen. This is entirely standard and reasonable, just something we've never done. I'll open another issue on that.

gordonwoodhull:
no, this is not a duplicate.  #665 is about having parallel areas and being able to move code easily between notebook and session.  this is still about running everything in the notebook, but being able to more easily edit the cells en masse.

shaneporter:
Rather than showing a border around the cell (which would contract the cell's contents), perhaps a semi-transparent background to indicate selection would be more attractive, and obvious, to the user?

![image](https://cloud.githubusercontent.com/assets/2493614/12643728/c206e5da-c5b5-11e5-9647-2ce952608e0a.png)

I'm also thinking that when the user selects 2 or more cells, it would be useful to show a panel that lists the selected cells by name. If the panel had a button, such as 'inverse selection', it would be easier to understand than 'crop', prior to a delete. Most obviously, a `DEL` key shortcut would do the job of deleting, but a 'Delete' button on the panel could do the same.

In the future, the panel could be developed further, so that users could print cells selectively.

gordonwoodhull:
I agree with the transparent background - I guess it is technically an overlay with `pointer-events: none`?

I'm not sure about the panel - it's always an issue of screen real estate. Even with the navbar buttons, which sound like a good idea, space is tight especially on laptops. In particular, I don't think we need a print button - for now, we can have print always print the selection if there is any, and we can consider custom print options or print preview with options later. 

[...]

Issue #300, opened 2014-01-27
document keyboard shortcuts


smschauhan:
There are a bunch of useful ace editor shortcuts which users aren't aware of. It would be helpful to have a list of useful shortcut help available. There are a bunch of them available [here](https://github.com/ajaxorg/ace/wiki/Default-Keyboard-Shortcuts)

gordonwoodhull:
Taking this for 1.6 as I don't have time for the current release. If anyone else wants to compile it earlier, obviously that great. IMO it should go on http://att.github.io/rcloud/ and we should link to it from the internal about page.

martymcwizard:
As a future enhancement, I thought that we could rename the help in the bottom left to be R help and add a new selection UI help and have some information there.  Just a thought.

gordonwoodhull:
Yeah, instead of the hint ".. Get help on" it could say "R help", and have a link to the right for _UI Shortcuts_. Much like we have a link for _Lucene search syntax_ up in Search, which perhaps should also show in the help pane instead of another tab.

shaneporter:
Correct me if I'm wrong, but as I understand it, the current (non-Ace editor) shortcuts are assigned in init.js. With the number of these set to grow, it makes sense to split this shortcut management into a separate object/file,`shortcut_manager.js`.

A shortcut could be defined with the following properties:

1. Shortcut code (note this would be undefined for something already handled, such as something within the Ace editor)
2. Category of the shortcut, such as 'Cell Editor', 'Cell management', 'Notebook Management')
3. Shortcut description, such as 'Deletes the selected cells'.
4. Function to invoke when the shortcut is performed (undefined when the shortcut code is also undefined)
5. Text representation of the shortcode. (`del`)

We could even define variations for Mac/Windows/Linux and present these shortcuts based on the client.

The `shortcut_manager` would set up the shortcuts (by code) where appropriate and a view would use the collection of shortcuts to produce the documentation for use within RCloud.

gordonwoodhull:
Yes - it looks like the only other relevant `keydown`s are in `ui/find_replace.js`

A global manager sounds like a good idea. Please use the (simple but effective) `RCloud.extension` framework for the basis - in particular the init/load distinction allows add-on packages like IIP and RCAP to manipulate the lists before they are realized in the browser. I guess the framework's sort order would be used for the documentation.

[...]

PR #1809, opened 2016-03-02
Feature/multi select cell

shaneporter:
Pull request for multi-select cell #658. Note that crop has been implemented as 'toggle selection -> delete'.


gordonwoodhull:
This works great and the code is very clear!

I wouldn't have thought of making the selection state part of the notebook model, but that makes a lot of sense.

The main suggestions I have are aesthetic:
 - the selection bar is logically part of the notebook and shouldn't blend into the title bars of the side panels. I appreciate that the current design creates clean lines, but here, combining the selection bar with the side panels confuses me. I think it should be a lighter grey (maybe somewhere between the cell commands and the panel titles).
 - the checkbox control can also be shorter IMO. This is the most crucial real estate so if the "outer button" can just fit the checkbox I think that would be a win.
 - notice the weird effect when the find and replace panel is open (ctrl-F) - I think this is the correct order (selection feels conceptually closer to the cells than find?) but it underscores that the selection bar is part of the notebook, not part of the navbar.
 - the equivalent control in gmail uses the entire outer button to trigger the dropdown. That saves one line of clutter, and is a sensible action for clicking outside the checkbox but inside the button (which currently does nothing).
 - the yellow is a bit violent - maybe we can use something more subtle, like a pale green? and maybe it should just fill the cell command bar and the gutter, not the code and results? I'm not sure. Right now it interferes with editing because it is so loud. I wouldn't want people to have to deselect just to continue editing code.

gordonwoodhull:
I also have a round of UI improvements we can consider after this is merged:

 - I notice that clicking the cell name toggles selection, which is great, but I think clicking the rest of the cell bar should do the same thing. (The cell name and the cell bar are the same object.)
 - I would expect clicking a cell name/cell bar to have a slightly different effect from clicking the checkbox. It should be more like the multiple selection interface: if the cell is already selected, it should have no effect. If it is not selected, then the selection is cleared and only that cell is selected.
 - We'll eventually have to consider if we need [multisortable](http://jsfiddle.net/neochief/KWeMM/) - maybe not for this release, but I think it's what people will expect, and it's also helpful, e.g. to move a bunch of currently dispersed cells to the beginning or end. Multisortable is not part of core jQuery UI, so I don't know if there is a stable implementation.

gordonwoodhull:
Great work!

htdocs/js/notebook/notebook_model.js
```diff
@@ -174,6 +174,35 @@ Notebook.create_model = function()
...
         },
+        remove_selected_cells: function() {
+            var that = this;
+            _.chain(this.cells)
+            .filter(function(cell) {
+                return cell.is_selected();
+            })
+            .each(function(cell) {
+                that.remove_cell(cell);
+            });
```
gordonwoodhull: It doesn't really matter, but underscore isn't needed here - `this.cells.filter(...).forEach(...)` is fine.

gordonwoodhull: (not that our code has any consistent style about this sort of thing either)

shaneporter:
image

This shows an updated selection bar (d222aa8, 69ced7f and ac2833e) and selected cell change (not committed at time of writing.) Not sure what I think about it so welcome a second opinion. In fact, I've just thought of an alternative, which I will add in another comment.

I haven't changed the colour, preferring not to change to green since this might interfere with the cell status (circled) which is also green.

I wasn't 100% about what was necessary for the selection bar, but have made it lighter. Visually, it is separate from the side panels' title bars. The checkbox selection area is now slimmer.

gordonwoodhull:
Yeah, we can continue to revise it, but I would consider each of those points fixed with the design you show above.

gordonwoodhull:
With the selection bar in that size and color, it enhances the illusion that the notebook is "above" the left and right columns. I like it.

shaneporter:
Hmm... that yellowish box shadow didn't really work across a range of cell sizes. For larger ones it was ok, but for smaller ones, it was rather overpowering.

So, I've adopted your suggestion of filling the command bar and gutter:

image

It's fairly unobtrusive, but obvious at the same time. (Committed with 8fa665a).

shaneporter:
One of your post-merge considerations, 'short cut for delete' had been implemented. 

Invert and crop/cull shortcuts shouldn't take too long to implement. Can you suggest their keyboard shortcuts?

gordonwoodhull:
That's a bigger question than it sounds - I think we need to look at shortcuts from a global level as part of #300.

As a placeholder, let's use what Preview uses on the Mac, ctrl/cmd-shift-I for invert, and ctrl/cmd-K for crop.

shaneporter:
Sounds good to me.

gordonwoodhull:
I'm not sure if I agree with myself about the bar and gutter (off balance?) but it's a good start. As you said, it's unobtrusive and obvious, so it's ready for people to test. 

Let's merge this tomorrow and then continue to iterate once we get some feedback from users.

gordonwoodhull:
Oops, I missed that final commit. Merging now.

BTW, we switched to grunt instead of make (nice!): please rebase any feature branches you're working on (or merge from develop).

gordonwoodhull:
Merged!

There are a lot of implications we'll have to work through, but I can tell this is going to be very, very helpful for folks. Thank you @shaneporter!

Eventually I'll open separate issues for some of the stuff discussed above. Let's get a little feedback first.

Issue #1789, opened 2016-01-27
modernize the build


gordonwoodhull:
Some of these topics are probably elsewhere, but I wanted to start a new issue because I'd like to tackle all of this at once.

We're still using make, which is ill-suited for building JavaScript and we're missing out on a lot of useful build tools because of it.

I'd like to 
 - [x] switch to Grunt
 - [x] concatenate the scripts with grunt-concat or something similar. we'll get source maps this way.
 - [x] uglify with grunt-uglify (even though we don't currently use the uglified version for debugging reasons - maybe source maps actually work by now)
 - [ ] use SASS to help organize the CSS (#1854)
 - [ ] use jscs and jshint for linting (lax, minimal) coding standards, and implement commit hooks (#1855)

All of this can be borrowed from [the dc.js Gruntfile](https://github.com/dc-js/dc.js/blob/develop/Gruntfile.js) where it's been helping for years.

gordonwoodhull:
Related: #1728 delete artifacts

gordonwoodhull:
I tried this on a branch and it worked immediately. Look, source maps!

![image](https://cloud.githubusercontent.com/assets/1366709/13479437/d815f988-e0a4-11e5-845e-e95c4072f4c0.png)

I'll open a PR because I'm not 100% sure of the consequences.

gordonwoodhull:
This is merged. Opening new issues for the remaining work.

PR #1810, opened 2016-03-02
Use grunt instead of make

gordonwoodhull:
and generate source maps

@s-u, I think we discussed using npm's grunt tools instead of make, mostly so that we can have source maps and don't have to search through rcloud_bundle.js all the time. 

as described in #1789, it will also help us clean up the CSS with SASS, and lint the JS.

does this look okay to you? if you take a look at the artifacts commit you can see that the result is exactly the same except for some line breaks.

it just means you have to `npm install` at root instead of in those two directories. build.sh is a little bit slower, too, but not significantly.


gordonwoodhull:
@s-u, are you okay with this one? It will help front-end debugging a lot.

s-u:
Yep, thanks, good with me.

s-u:
Actually, the `build.sh` doesn't seem right:

```
$ sh scripts/build.sh 
scripts/build.sh: line 47: grunt: command not found
``

Obviously, there is no `grunt` command (it's a module not a system binary) so that probably needs fixing ...

**Comment by gordonwoodhull** [comment]
Hmm, good point. 

We require `npm` to be installed and this would assume that you also installed grunt globally using `npm install -g grunt-cli `

But we can eliminate that step and run `node_modules/grunt-cli/bin/grunt` instead, I think. Will change & test this before I merge.

**Comment by s-u** [comment]
After some googling it seems that it also requires `grunt-cli` so you have to add that to `package.json` and then

```
diff --git a/scripts/build.sh b/scripts/build.sh
index b24e32c..f4021d6 100755
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -44,7 +44,7 @@ fi
 
 # build JS (if available)
 if [ -d ./node_modules ]; then
-    grunt
+    node_modules/grunt-cli/bin/grunt
 fi
 
 # check if we need to worry about mathjax
```

gordonwoodhull:
Merged!

Issue #1781, opened 2016-01-19
load notebook tree more asynchronously


gordonwoodhull:
This is another follow-up to #405, first load is slow. 

Previously we made building the tree UI lazy, but we're still loading all the data, and this is slow, both because it's a huge RCS query and because so much data must be crammed through the wire.

The more intelligent thing to do would be to model the tree on the back-end. Currently the server has no idea that there is a notebook tree, so this is is some work, but it might help with stuff like #1402 reflecting changes to the tree made in other sessions.

The easier thing to do would be to load each user's notebook tree on-demand. Since we currently don't have any features that organize the notebooks in any other structure, this would be fast and effective.

gordonwoodhull:
Actually, we don't need to model the whole tree on the server. All we really need to do is load the notebooks for each user on demand. We estimate that the most active users of RCloud will have 500-1000 notebooks, which should be an acceptable amount of data to load. The problem arises with hundreds of users times hundreds of notebooks.

We only open one branch of the tree at once, so we only need to load one user's notebooks at a time. Then we can build that tree and display it.

The relevant API is `user.all.notebooks`, which is not currently exposed on the client. I'll expose that and then @shaneporter, you should be able to rework the editor_tab to load parts of the tree on demand. 

https://github.com/att/rcloud/blob/develop/rcloud.support/R/rcloud.support.R#L558

This fix is high priority for this release.

gordonwoodhull:
Done - the JS API is `rcloud.config.all_user_notebooks(username)`

gordonwoodhull:
Implemented in #1901

gordonwoodhull:
Here are the improvements for me (couple hundred starred notebooks), on a fully-populated but underutilized develop server (rcloud-5).

### Without this optimization
Notebook infos loaded: **11483**
Average first load time: **16.92s**
Time spent initializing UI: ~**2.4s**

### With lazy loading
Notebook infos loaded: **316**
Average first load time: **4.88s**
Time spent initializing UI: ~**0.8s**

shaneporter:
Pretty good. 🙂

Issue #1780, opened 2016-01-19
load notebook before tree?


gordonwoodhull:
Starting an issue for this idea, originally mentioned in "first load is slow"

https://github.com/att/rcloud/issues/405#issuecomment-39006032

The load would feel faster if the notebook loaded before the notebook tree. Even if we load the tree more asychronously, it could still help to load the notebook first.

gordonwoodhull:
I started experimenting with this and some of the cases are not too hard, although it tangles up the initialization promises further.

Specifically, there are three cases: load a particular notebook, load last accessible notebook, and create new notebook. The first two cases should be okay, although it does mean delaying the update to the tree until there is a tree to update. 

Interestingly, the third case does require the notebook tree to be loaded first, because a new notebook has to look at the tree to determine the name of the new notebook. If we had a server-side model of the tree as suggested in #1781, but for now we'd have to handle this case differently, and create new notebooks only after the tree is loaded. 

Since loading existing notebooks is the more common case (new notebooks are usually created from an open RCloud page), it's probably okay to only handle those cases.

gordonwoodhull:
This is a pretty big refactor, because most of the editor_tab assumes that the notebook tree and the "notebook registry" will be loaded before any notebooks are loaded. It's worth doing, but it's not a simple tweak, because there are many cases involved in notebook loading.

It might make sense to tackle this at the same time as #1781, and really seriously untangle this code all at once.

gordonwoodhull:
Interestingly, while the notebook tree still starts loading before the notebook, lazy loading of the tree #1781 now means that the notebook shows up first. This can still be improved, and lazy loading has set the stage to allow loading the notebook even earlier.

Issue #1075, opened 2014-11-10
sort all notebooks by date


gordonwoodhull:
This is more difficult than #522 sort user's notebooks by date, because it means there's no tree, and there's an awful lot of data to display.  

To make it more manageable for users, we could keep the same categories (RCloud Sample Notebooks, Notebooks I Starred, People I Starred, All Notebooks), but under each there is just a flat list of notebooks, so still a tree but one level less of it.

Issue #522, opened 2014-04-11
sort user's notebooks by date


s-u:
It would be useful to be able to sort notebooks by the commit date instead of name.

gordonwoodhull:
Implemented in RCloud 1.9

s-u:
Great! It actually works very well by sorting the folders by the most recent date of their content.

Minor complaint: it takes ~10s to update on my notebooks - both ways (date->name and name->data)

gordonwoodhull:
Odd, it only takes 5s on my machine when your notebook tree is open. We can continue to optimize though.

PR #1842, opened 2016-03-15
Feature/document shortcuts

shaneporter:
Initial changes for documenting shortcuts. Try '?' to see the shortcuts dialog.

Points of interest:

- the shortcuts are managed using [Mousetrap](https://craig.is/killing/mice)
- `shortcut_manager` uses the extension API. Not sure if it's using it in the intended way, but I welcome comments.
- shortcuts are added using the following code:

```
 RCloud.UI.shortcut_manager.add([{
        category: 'General',
        id: 'show_help',
        description: 'Show shortcuts help',
        keys: [
            ['?']
        ],
        action: function() { RCloud.UI.shortcut_dialog.show(); } 
    }]);
```

the `keys` property is an array of arrays; let's have a look at another example:

```
 keys: [
    ['ctrl', 'k'],
    ['command', 'k']
 ],
```

This contains options for non-mac and mac. The shortcut manager filters out mac commands for non-mac systems using the `ui_utils.is_a_mac' function. 

On my PC, I don't see the 'command k' option.

- Prior to registering, shortcuts are checked to see if they clash. If so, they are not registered, and a warning message output to the console.
- The category property, which defaults to 'General' if not supplied, is used to dynamically construct the shortcuts dialog's headings.


gordonwoodhull:
Excellent. Good thought to use a library for key bindings. I look forward to reviewing this! 

(Unfortunately I'm out with the flu, so that may not be until later this week.)

shaneporter:
Sorry to hear you're not feeling well. I've just done the cell select related enhancements and submitted a pull request for those.

What would you like me to focus on next? I'm thinking of taking a look at the Notebook recommendations. I don't have anything against AngularJS, but it seems a little overkill for what's currently there. I could definitely trim things down a lot.

gordonwoodhull:
Yes, cleaning up the recommendation code sounds like a good idea.

If you're up for a bigger challenge, we have a cluster of related requests around sorting the notebook tree - #522 #1075 and some other linked issues. This code (in editor_tab.js don't ask) is the oldest and badly needs a refactor. It's not on our list for this release but we have to attack it sometime. 

It's also the source of our biggest performance bottleneck, which is in this release, but solving that may require some server side work as well.

gordonwoodhull:
The perf problems are #405 #1780 #1781 - even if you don't take this on, another pair of eyes will help.

shaneporter:
I have pushed a fix for issue #1847 with this pull request (since it overhauls the shortcuts management.)

gordonwoodhull:
This looks great. My only suggestion is that it only display the shortcut keys for the current platform, since no one is going to care about other platforms and it just clutters it up. I'll open a separate issue for that.

Now we just have to add a bazillion other shortcut keys!

Issue #1834, opened 2016-03-11
Check box should be disabled, if there are no cells present


Prateek032:
User can select the check box for once even if there are no cells present in the notebook

shaneporter:
I'd say we can go further and disable the check box, drop down (all, none, invert) and delete button.

gordonwoodhull:
My preference is for the checkbox also to be in the mixed or non-state (horizontal line) since it's indeterminate.

I don't think we currently have any logic that detects whether there are cells or not.

shaneporter:
I've thought about this a bit more, and don't think we should use a non-state, since that is used specifically for 'some but not all selected'. In the case where there are no items, and hence no selections, the unchecked state should be used.

![image](https://cloud.githubusercontent.com/assets/2493614/13751648/0cc40a82-ea03-11e5-8aa7-feace85b6c63.png)

Gmail uses an unchecked state checkbox when there are no items... and doesn't disable the clicking of the control to show the menu. Clicking on a menu item or the checkbox directly has no impact and the checkbox remains in an unchecked state.

gordonwoodhull:
Okay, sounds good. I was probably being pedantic. A disabled empty checkbox will probably be more clear to most folks.

gordonwoodhull:
fixed in #1843

Issue #1826, opened 2016-03-10
select single for unmodified click on cell title/bar


gordonwoodhull:
Cell title/bar selection should behave slightly different from checkbox selection.

Currently clicking on a cell title/bar has the same effect as clicking the checkbox, toggling the selection state of that cell. Instead, if no modifier key is used, only that cell should be selected, and all other cells should be unselected.

It should work mostly like multiple selection in a file browser.

gordonwoodhull:
fixed by #1843

Issue #1825, opened 2016-03-10
click on cell titlebar to select


gordonwoodhull:
Clicking on the rest of the bar should have the same effect as clicking on the cell name, just as either one can be used to drag the cell.

In particular, one is currently forced to use only the checkbox if "Show Cell Numbers" is disabled.

gordonwoodhull:
fixed in #1843

PR #1843, opened 2016-03-15
Enhancement/select cell improvements

shaneporter:
Proposed fixes for:

- #1834 (though this does behave like Gmail, and does not disable)
- #1827 
- #1826 
- #1825


gordonwoodhull:
This is very close. However there are still a few quirks:

Re #1834 no cells:
 - There should be a visual indication that the checkbox and the menu items are disabled, i.e. they should be greyed out. (I notice gmail doesn't do this, but Google isn't particularly known for its UI expertise.)

Re #1826:
 - Although unmodified click now correctly selects just one, ctrl/cmd-click should still toggle the selection when clicking the title/bar. This is consistent with file browsers. This is one of those places where we have to special case OSX and separate ctrl- and cmd-, because ctrl-click means right click on OSX.
 - There's now a bug with clicking a checkbox where it selects properly but the checkbox itself can get out of sync (the cell is selected but the checkbox is not, or vice versa). Perhaps the event is being handled twice, once for the bar and once for the checkbox?

shaneporter:
@gordonwoodhull another update; can't check the first point regarding #1826, since I don't have a mac, but hopefully it will now work as expected.

gordonwoodhull:
Looks like everything is fixed but single selection doesn't work anymore. So I can toggle but the behavior where clicking on the title/bar of an individual cell does not select that and deselect everything else.

I think I'll go ahead and merge this and open a new issue, since this is a lot better.

gordonwoodhull:
merged

PR #1844, opened 2016-03-16
Preliminary tidy-up of 'discover' code

shaneporter:
Hi @gordonwoodhull,

I know there are plans to further develop the 'discover' area of RCloud, but having taken a look at the code, I felt it was a good opportunity to slim things down a little.

The end result should be exactly the same, but there's a lot less code in there, and the Angular dependency has been removed.

Thanks,
Shane.


htdocs/js/ui/discovery_page.js
```diff
@@ -1,37 +1,81 @@
...
+                          image_url: 'notebook.R/' + notebook[0] + '/thumb.png'
+                        }
+                      })
+                      .value();
+
+                      $('progress').attr({
+                        max: recent_notebooks.length
+                      });
+
+                      var template = _.template(
```
gordonwoodhull: cool, i didn't know about `_.template`!

gordonwoodhull:
Nice progress element. (Hopefully that will be a lot faster once we stop pulling thumbnails from github.)

gordonwoodhull:
merged!

Issue #1637, opened 2015-08-18
notebook propover title spills outside of the box, when a name is a long single word


anatoliyg:
it looks like the truncation mechanism doesnt kick in when the title is a single long word.
<img width="902" alt="screen shot 2015-08-18 at 1 39 14 am" src="https://cloud.githubusercontent.com/assets/1855548/9332177/68b253d8-4591-11e5-8d83-129b846a03c0.png">

anatoliyg:
duplicate of https://github.com/att/rcloud/issues/1593

anatoliyg:
fixed by e5e0e8e

Prateek032:
We were able to reproduce this issue again

![nb_name](https://cloud.githubusercontent.com/assets/10941328/13802984/73c1b486-eafe-11e5-966f-4b3757afe31e.PNG)



![nb_name1](https://cloud.githubusercontent.com/assets/10941328/13802985/73c4c4c8-eafe-11e5-90ec-43f0ee014b99.PNG)

gordonwoodhull:
Yay

shaneporter:
Hi @Prateek032 - how do I show the 'Notebook Permissions / Group Management' dialog?

shaneporter:
This happens on the 'discover' page too:

![image](https://cloud.githubusercontent.com/assets/2493614/13808513/4734da64-eb5e-11e5-9d02-d5b515ace3fe.png)

gordonwoodhull:
@shaneporter, the notebook permissions dialog is accessible through the notebook info popover - just click on the group link there. 

It's not very well advertised and we've talked about merging it with the notebook public/private command (the eye) but currently those are two unrelated functions (#1549).

shaneporter:
thanks @gordonwoodhull, but on my local machine, all my groups are listed as static text _public_.

gordonwoodhull:
@shaneporter, ah right, of course. You'll need to enable notebook protections. 

Follow the instructions in services/README.md for SessionKeyServer, and add the following line to rcloud.conf:

```
Session.server: http://127.0.0.1:4301
```

This requires Java. Basically it's a simple isolated server to create and provide keys for encryption (and user switching, but we rarely use that in testing).

shaneporter:
(Incidentally, another commit, f65c4f5, should fix the issue on the discover page)

gordonwoodhull:
confirmed that discover page is fixed by #1844

gordonwoodhull:
[...]

PR #1853, opened 2016-03-17
proposed fix for #1637

shaneporter:


gordonwoodhull:
merged. thanks @shaneporter!

Issue #1852, opened 2016-03-17
Firefox: Notebook name is not getting wrapped in firefox browser


Prateek032:
If notebook has long name,  in chrome browser we see text is getting wrapped but in firefox  it is not 

![nb_ff](https://cloud.githubusercontent.com/assets/10941328/13837441/ad8420e4-ebc8-11e5-86b9-33967cc497b5.PNG)

shaneporter:
This can be fixed by adding `display: inline-block;` to the item's CSS selector, which overrides the default (inline, since it's a span element) in `jqtree.css`. 

Is it a good idea to create a `jqtree-rcloud-tweaks.css` file and place it in there?

gordonwoodhull:
I think it's fine to add it to bootstrap-rcloud-tweaks.css (since that should be named rcloud.css) or edit.css. Want to take this?

shaneporter:
Yes please.

gordonwoodhull:
Thanks!

shaneporter:
Proposed a fix with #1856

gordonwoodhull:
fixed by #1856

PR #1856, opened 2016-03-17
Proposed fix for #1852

shaneporter:
Fix for #1852, with the caveat that the info is displaced for long notebook titles.


gordonwoodhull:
merged, thanks!

Issue #1848, opened 2016-03-17
Cell selection color should be consistent


Prateek032:
Currently we have these two behaviors
 For editable cells selection looks like 

![cell_edit](https://cloud.githubusercontent.com/assets/10941328/13837053/9dade32a-ebc4-11e5-8c10-fb9163812e0d.PNG)

and for non-editable cells it look like
![cell_nonedit](https://cloud.githubusercontent.com/assets/10941328/13837065/b4cb8ce2-ebc4-11e5-8893-3caeeaab3784.PNG)

gordonwoodhull:
Yeah, the gutter should be yellow when the cell is active, as well. Which I suppose means the "active line shadow" becomes a dark yellow.

gordonwoodhull:
fixed by #1857

PR #1857, opened 2016-03-18
fix for #1848

shaneporter:
Suggested fix for #1848; it looks like this:

image

Cheers, Shane.


gordonwoodhull:
perfect, just perfect

PR #1865, opened 2016-03-22
Shortcut related fixes

shaneporter:
Please review. Thanks.


gordonwoodhull:
An odd side-effect of this change is that where pressing ctrl/cmd-F a second time used to do nothing, now it opens the browser's internal find dialog. I'll open a new issue for that.

gordonwoodhull:
Merged. thanks @shaneporter!

PR #1866, opened 2016-03-22
fix for firefox long names

shaneporter:
introduced `word-break:break-all` into the mix:

image

image


gordonwoodhull:
Merged, thanks @shaneporter!

PR #1868, opened 2016-03-23
Enhancement/sass

### LINKED ISSUE #1854: use SASS
Follow up to #1789.

Now that we are on Grunt, we can update some other tools easily.

Let's use SASS. It's succinct and it will give us an opportunity to organize and make sense out of the CSS, which is a sprawling mess.

The general-purpose CSS is in the historically named bootstrap-rcloud-tweaks.css, and stuff that is needed only for the edit or view mode are in the respective .css. Open to suggestions, but I think the first should be called rcloud.css and the others should be called rcloud-edit.css and rcloud-view.css.

This replaces #88.

shaneporter:
fixes #1854 

The end result is four individual CSS files: 

- `rcloud.css` - base file used by `edit.html`, `view.html` and `mini.html`
- `rcloud-discover.css` - used by `discover.html`
- `rcloud-edit.css` - used by `edit.html`
- `rcloud-view.css` - used by `view.html`

`mini.html` and `goodbye.R` use the base file only.

There are accompanying maps for each of these files.  

The base file imports a number of other `scss` files, which are copies of the original `css` files, moved into sub-directories of `sass`. Output is written into the `css` folder. 

I have also updated the `INSTALL.md` to mention SASS.

Incidentally, SASS v3.4.0 is required to generate source maps. (I had 3.2.12 on my machine and wondered why it wasn't working.)


gordonwoodhull:
Looks like we can use `grunt-sass` instead of `grunt-contrib-sass` to eliminate the need to install sass separately, as used in this PR on another project of mine: https://github.com/dc-js/dc.js/pull/1049

We're trying to avoid having build dependencies that need to be installed separately, aside from Node itself.

gordonwoodhull:
The other half of this work, which can be done after we verify this, is to organize all the SCSS hierarchically. That's the real win IMO, because right now the rules are completely disorganized and in in random order. But it might be very tricky.

gordonwoodhull:
Looks like the only problem with `grunt-sass` is that it doesn't offer compression built in (?) 

So we might have to add `grunt-contrib-cssmin` or something. (Although currently we don't use minimized anything out of perhaps outdated debugging concerns.)

gordonwoodhull:
My rebase-branch is `merge1868`, in case I don't merge it soon. I just realized this is totally not what I should be working on... we have a patch release coming up this week and there is no merge order that will avoid a difficult merge between the css changes I need to make there, and this.

gordonwoodhull:
Not merging this yet, just because it's a messy merge.

gordonwoodhull:
I rebased this and merged it before things diverged any further. (Any continuing changes to the css do not merge with these renames, so I applied them in another commit.)

Issue #1873, opened 2016-03-24
Error when <DEL> key is pressed in anonymous view.html


Prateek032:
Open any notebook in 'view.html' and press < DEL > key, just below the cells following error is displayed

![del_err](https://cloud.githubusercontent.com/assets/10941328/14011627/432eed1a-f15d-11e5-866e-522b2d7a7f33.PNG)

The same error is displayed for the Read only notebooks as well. Just open any notebook in read only mode and press < DEL > key

gordonwoodhull:
Thanks @Prateek032!

gordonwoodhull:
fixed by #1878

gordonwoodhull:
Now for anonymous user.

shaneporter:
I can't reproduce this. This is what I'm doing:

- publish notebook
- log out
- go to view page for the published notebook (I see '--- welcome, anonymous user' at the top of the notebook)
- press <del>

For me, nothing happens. Nothing in the console.

Prateek032:
@shaneporter , Try without publishing the notebook and we got the following error

![image](https://cloud.githubusercontent.com/assets/10941328/16140822/4ff92972-3472-11e6-9dd5-89c714f4d8ba.png)

shaneporter:
thanks @Prateek032, but I still can't reproduce locally. If you show the shortcuts dialog with `?`, you're probably not seeing much because of #2083, but locally I see that `del` isn't registered as a shortcut for 'view', so when I press `del`, nothing happens.

gordonwoodhull:
Right now I see this only when the notebook is not published and the "Could not load notebook. You may need to login to see it." message appears. 

Apparently all shortcuts are enabled in that case!
![image](https://cloud.githubusercontent.com/assets/1366709/16156358/e04a3700-3481-11e6-90a2-db7c6705e761.png)

Not that it matters, but I see a different error from either of the ones shown above
![image](https://cloud.githubusercontent.com/assets/1366709/16156397/165c6b56-3482-11e6-94aa-f759bc276879.png)

shaneporter:
I see that now @gordonwoodhull. Taking a look.

PR #1878, opened 2016-03-24
categorize shortcut keys by readonly

shaneporter:
I use `shell.notebook.model.read_only()` to interpret the meaning of the modes.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -113,6 +113,7 @@ RCloud.UI.init = function() {
             ['command', 's'],
             ['ctrl', 's']
         ],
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Thanks @shaneporter. Please title PRs with something human readable. :wink:

gordonwoodhull:
I think this addresses #1873 as well.

gordonwoodhull:
I was hoping that the existing category functionality in RCloud.extension would work here, but this works too, and I appreciate that it's consistent with the interface.

gordonwoodhull:
Merged, thanks @shaneporter!

PR #1881, opened 2016-03-24
provide shortcuts for current platform only

### LINKED ISSUE #1861: only display shortcuts for current platform?
Opening this for discussion, since I initially thought it was a bug but it's actually correct but confusing.

Seeing keyboard shortcuts for other platforms may overwhelm users. (Especially windows users going "what's the command key?!" ;) 

We generally map everything to both ctrl- and cmd- for simplicity of implementation. But maybe the shortcuts dialog should only show the shortcuts for the current platform?

shaneporter:
If I spoof a Mac, this seems to work, so hopefully the issue can be closed.

### DIFF
--- a/htdocs/js/ui/shortcut_manager.js
+++ b/htdocs/js/ui/shortcut_manager.js
@@ -27,8 +27,10 @@ RCloud.UI.shortcut_manager = (function() {
                     return _.contains(keys, 'command');
                 });
             } else {
-                // this is a mac; if this shortcut has a command AND a ctrl, remove the ctrl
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Okay, I feel like this might have weird effects if we ever have a shortcut for Mac that does use the control key, but so far I don't think that's an issue.

gordonwoodhull:
Ahh, this looks much more reasonable:

image

gordonwoodhull:
Merged, thanks @shaneporter!

PR #1890, opened 2016-04-05
Fix for first time cell click scroll/jump badness.

### LINKED ISSUE #1624: clicking on cell code the first time causes horrible scrolling effects
I think there are probably half a dozen bugs here. The window scrolls, the code scrolls, the cursor ends up in a different place from what was clicked. 

It is hard even to diagnose the problem there is so much wrongness here.

Users report this is worse as of 1.4, so perhaps the fix for #1358 exacerbated it somehow.. That commit was 48122f6aead5ac72434dcd042f9a41b798fea6f3

But we aren't sure, it might be that users just remember things working before RCloud 1.2 changed the cell implementation from Ace everywhere to static text that must be activated.

shaneporter:
This pull request should mean that:

- Cell height stays the same when it goes into edit mode.
- Main scroll position stays the same.
- Cursor position should be where you clicked (however, because the font changes when it goes into edit mode, there's sometimes a slight difference in position of caret relative to characters.)

(fixes #1624)

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -11,6 +11,7 @@ function ensure_image_has_hash(img)
 
 var MIN_LINES = 2;
 var EXTRA_HEIGHT_SOURCE = 2, EXTRA_HEIGHT_INPUT = 10;
+var scrollable_area = $('#rcloud-cellarea');
 
 function create_cell_html_view(language, cell_model) {
     var ace_widget_;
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Neat. I thought that we couldn't have ace cells taller than the window, but I wanted to try this approach of keeping them at full height. 

It works! No more double-scrolling!

I'm not seeing the font change; I think it's easier to match the font on OSX than on Windows. I am seeing slight differences in how word wrapping happens, but I don't think we can do anything about that because Ace seems to implement its own special word-wrapping beyond what CSS has.

gordonwoodhull:
Rebased and merged to develop

gordonwoodhull:
Thanks @shaneporter!

PR #1895, opened 2016-04-11
Added crop button to selection toolbar

### LINKED ISSUE #1871: crop on selection bar
since we offer a keyboard shortcut and trash, lets add a button as well.

We tend to bend the meaning of icons a little bit, so it's probably okay to use the standard crop icon even though this has nothing to do with rectangular selection

image

http://fortawesome.github.io/Font-Awesome/icon/crop/

shaneporter:
Added button to toolbar and hooked up its click event to the existing notebook controller's `crop_cells()` method.

This fixes #1871.

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -146,6 +146,7 @@
       </ul>
     </div>
     <button id="selection-bar-delete" title="Delete selected cells" type="button" class="btn btn-link navbar-btn"><i class="icon-trash"></i></button>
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @shaneporter. I put it first because.. I don't know why. And added a little padding.

It occurs to me belatedly that this button is rather dangerous, because if nothing is selected, then it deletes everything. We might need a warning in that case, guess I'll open an issue.

Issue #1863, opened 2016-03-21
esc to close keyboard shortcuts dialog


gordonwoodhull:
most of our dialogs close with `<esc>` - where they don't, it's a bug (#1831)

shaneporter:
I've added a global handler for `<esc>`. I'm not sure if this is overkill, since you mention that

> most of our dialogs close with `<esc>`

If it's required that they do, this should fix that issue too. The selector '.modal' is used, which, I'm assuming is used for all bootstrap dialogs.

gordonwoodhull:
hmm, that is kind of overkill, but it works. let's see if it causes problems heheh.

gordonwoodhull:
fixed by #1865

Prateek032:
This issue still persists on Local instance

gordonwoodhull:
Thanks @Prateek032, reopening.

shaneporter:
This has been fixed.

gordonwoodhull:
Fixed by #1896

Issue #1889, opened 2016-04-04
browser find pushes through on second cmd-F


gordonwoodhull:
I presume this also happens with ctrl-F on Windows.

Prior to #1865, pressing the find key a second time would be ignored. I guess we are grabbing the keys in a more polite way.

(I actually like the behavior where it cancels find or switches to replace, but that's kind of complicated and non-standard, so I took that out in favor of `esc` at some point.)

Now it opens the browser's find dialog, which is confusing. Also, cmd-opt-F goes to the location bar and enters a question mark, wasn't aware of that browser feature.

gordonwoodhull:
Fixed by #1896

gordonwoodhull:
Seeing this again, guessing it will also be fixed by #1919

gordonwoodhull:
fixed by #1919

Issue #1883, opened 2016-03-29
Inappropriate behavior of < ctrl + s > keyboard shortcut 


Prateek032:
Fails to save the cell content changes when we press < ctrl + s >. Instead of saving current changes, it is invoking 'File saving" window

gordonwoodhull:
Thanks @Prateek032!

shaneporter:
The same can be said for find/replace.

PR #1896, opened 2016-04-11
Bug fixes for browser find on find shortcut invocation and save.

shaneporter:
Save should now work from within a notebook cell.
Find/replace should work, even if invoked multiple times, and not cause browser-level shortcut handling for the equivalent shortcut codes.

Fix for #1889.
Fix for #1883.


htdocs/js/ui/shortcut_manager.js
```diff
@@ -21,7 +21,8 @@ RCloud.UI.shortcut_manager = (function() {
             var can_add = true;
 
             var shortcut_to_add = _.defaults(shortcut, {
-                category: 'General'
+                category: 'General',
+                modes: ['writeable', 'readonly']
```
shaneporter: This fixes #1863.

gordonwoodhull:

Issue #1867, opened 2016-03-22
open sample notebooks directly


gordonwoodhull:
RCloud Sample Notebooks is a folder which opens to a list of one user with the sample notebooks.

It should open direct to the sample notebooks.

gordonwoodhull:
When we create most of the roots of the notebook tree, we are duplicating All Users and changing the root node. 

I'm implementing this in the following way: if there is exactly one "featured user" then when the tree is duplicated, the user part is spliced out.

gordonwoodhull:
This is a pretty wretched hack, actually, since we assume pretty much everywhere that the notebook tree will have the form root/user/notebook.

But my hack seems to work, so let's see if QA finds any problems with it!

PR #1901, opened 2016-04-14
load notebook tree more asynchronously

shaneporter:
Hopeful that this will improve notebook tree load performance.


gordonwoodhull:
First issue: This does not work with the "featured users" feature. What's that feature, you ask. It adds another root of the tree at the top which shows "RCloud Sample Notebooks". You can see it in action on http://rcloud.social

To enable this feature, add a line to your rcloud.conf setting `rcs.system.config.featured_users` to a user whose notebooks you'd like to see in this section of the tree. If you haven't logged in as that user, do so and RCloud will create a Notebook 1 for that user.

Once you have that set up, you'll see an error where it tries to copy that user's notebooks to the special tree but the source tree hasn't been loaded yet. Your choice if you want to preload that user's notebooks, or lazy-load this tree in some way.

gordonwoodhull:
Second issue: If your current notebook is not yours, then this causes no notebook to be selected in the tree. However this is somewhat difficult to diagnose given #1905 where notebooks are often selected in the wrong tree, so I'm going assign you that issue as well. :wink: I think you brought this up on our call.

gordonwoodhull:
Note: all the notebook tree code is bad, but the RCloud Sample Notebooks code is the worst of the worst, because it's the feature which managers always focus on at the last minute because it's in the upper-left corner of the screen. So it's hacks upon hacks, always executed 2 days before the release.

shaneporter:
So the RCloud Sample Notebooks concept is so a specific user can login, create some 'showcase notebooks' to show off RCloud, a config setting points to that user, and that user's notebooks then appear in everybody's tree under 'RCloud Sample Notebooks'?

gordonwoodhull:
Yes, that's it. Underneath we allow multiple "featured users" but in practice we have a single "rclouddocs" user, and the UI eliminates one level of the hierarchy in this case.

htdocs/editor_tab.js
```diff
@@ -1306,7 +1306,7 @@ var editor = function () {
                 window.open(url, "_blank");
             }
             else
-                this.load_notebook(gistname, version, source, null, selroot);
+                this.load_notebook(gistname, version, source, selroot, null);
```
gordonwoodhull: ugh, too many arguments here. thanks!

shaneporter: no problem - I'm still looking at some very odd behaviour in this area, so not fully ready yet.

shaneporter:
Hi @gordonwoodhull - I was having an issue with featured notebooks not being selected correctly. I am now loading these at the start.

gordonwoodhull:
Breaking the notebook tree code into more functions helps the clarity a lot!

It looks like the new strategy is to load the metadata for the current user's notebooks, plus all their starred notebooks, plus the featured notebooks. I think this could be lazier but let's go with it as it should solve the performance problems for now.

However, there are a few bugs:
 - [ ] loading a notebook belonging to another users causes the notebook to be displayed as "hidden" (greyed)
 - [ ] loading sample notebooks causes duplication of the tree. (Probably it is mismatching ids and recreating trees that already exist. I ran into this problem recently when I implemented the skipping of the featured user level; you'll find those commits under "open sample notebooks directly" #1867.)
 - [ ] when loading the page, if the current notebook belongs to someone else, it is not highlighted (bolded). This works okay if the current notebook belongs to the current user, or if later loading a notebook belonging to someone else (it's grey but it's bold-grey), so it's probably a concurrency issue.

first two issues shown below
image

shaneporter:
Hi @gordonwoodhull - can you take another look at this please? It seems to be behaving itself a lot better.

gordonwoodhull:
Hi @shaneporter, I'm a little confused. I don't see any of the above issues fixed, except the first one is mostly better. (It usually shows the current notebook in black, not grey, when it belongs to someone else, although it is occasionally still grey.)

I hope my descriptions were clear. Let me explain a little further. (I don't think these are new, just describing the above issues in more detail.)

It's consistently duplicating notebooks in the tree when they are opened and belong to someone else:
image

If the notebook is in a folder then the folder gets duplicated:
image

This is not just sample notebooks as I wrote above. 

When the current notebook belongs to someone else and the page is reloaded, no notebook is highlighted (although that user's notebooks are displayed):

image

If that notebook is in a folder, that folder is also not opened. (So it's like the `selroot` stuff never happens. Again, I think this is probably because the lazy loading and highlighting is out of order.)

shaneporter:
Thanks for the clarification - perhaps I missed a commit or something - pretty sure I wasn't seeing the duplication issue. At a computer soon so I will investigate as a priority.

shaneporter:
Strange. I _was_ seeing duplication when opening another user's notebook (either at load time, or clicking on the tree), but I'm not seeing that locally now (because of a change I made yesterday.)

Nodes in a folder were not being automatically selected, but I have fixed that in 78ecbca.

I'm also not able to reproduce the greying nodes - 5ff2229 seemed to resolve that for me.

I'll continue looking into it. I suspect timing issues.

gordonwoodhull:
Okay, if you're not seeing these issues then I guess it is on me to debug it. I just deleted my local branch and checked out the code again, and I'm still seeing the issues. Thanks for verifying.

shaneporter:
Oh dear. I'm sorry you're experiencing issues.

gordonwoodhull:
I'm going to check into this later in the day, and also try another machine in case that matters. I should probably understand the new code better anyway! I guess you can get back to the other reviewed PRs.

gordonwoodhull:
Yes, the notebook tree probably needs a rewrite. Maybe we will find time for that later in the year.

gordonwoodhull:
Not seeing the issues on my other computer, very strange!

shaneporter:
:confused:

gordonwoodhull:
Yeah, I don't know whether that's a good sign or a bad one. At least I've got both in one place to narrow in on this.

shaneporter:
Good that you can reproduce in one place - I wasn't able to reproduce the issues at all after I made some changes.

gordonwoodhull:
@shaneporter, I added a secret option to show debugging tips for the notebook tree; [see this notebook](https://rcloud.social/edit.html?notebook=c29b46b80d8fd96376339b71a6176970) for the way to enable it on your local.

When you hover over a node in the tree, it will show the full path. And it seems like these bugs are due to the path being wrong. No idea why they weren't showing up on my work machine.

shaneporter:
Thanks @gordonwoodhull!

gordonwoodhull:
just noticed the distinct bug 
- [ ] loading a featured notebook displays as hidden (@shaneporter already fixed this for other users)

shaneporter:
Is that _the_ bug?

gordonwoodhull:
I've fixed two, that one is still to go.

It helps to have the right tools! It would be really difficult to write unit tests for this but just a little UI help makes all the difference.

gordonwoodhull:
So yeah, these all turned out to be the most minor glitches. @shaneporter, it's too bad that code is so scary, certainly your changes helped a lot!

gordonwoodhull:
Rebased and merged. Thanks so much!

shaneporter:
Thank you too @gordonwoodhull

PR #1903, opened 2016-04-15
Enhancement/ace shortcuts

shaneporter:
A few things of note:

1 - Modified the structure of shortcuts. The logic I was using previously didn't support mac/windows only shortcuts.
2 - Put the ace shortcuts into a separate file because there are so many of them.
3 - Didn't include all the shortcuts on the ace [default keyboard shortcuts page](https://github.com/ajaxorg/ace/wiki/Default-Keyboard-Shortcuts) because some of them didn't seem to be active for me (Multicursor, enter full screen etc...)

Cheers.


gordonwoodhull:
I'm getting 

```
Cannot read property 'join' of undefined(anonymous function) @ shortcut_dialog.js:45
```

gordonwoodhull:
This will close #300. We still need to add a ton more shortcuts but those are in separate issues.

shaneporter:
Hi @gordonwoodhull - please could you have another look at this? I have fixed the `join` error you saw, and the shortcuts seem to be working ok. Thanks.

gordonwoodhull:
Wow. That's an impressive list. Yes, it will probably need to be tabified.

I still get the `.join` crash for `ace_redo` but I'm checking in a workaround for now, so that we can go ahead with cutting the release branch. Will open another issue.

gordonwoodhull:
While merging this, I wasn't sure what to do about `ignore_clash`. You removed this checking entirely in #2000, and that's already been merged. But this PR introduced the flag, for I think the same purpose. 

So I think I've restored the checking correctly along with this flag, but I might have wavered so we'll have to test this area especially.

gordonwoodhull:
Rebased and merged. Can't guarantee I didn't introduce some bugs because it was a pretty hellish merge, so we'll have to test the keys thoroughly.

Thanks @shaneporter! We're ready to cut the release branch now!

Issue #1904, opened 2016-04-15
disable trash and crop buttons in selection bar when no cells are selected


gordonwoodhull:
We disable trash when there are no cells, but not when there is no selection. It has no effect, so it's harmless right now.

However, the crop button will delete all cells if none are selected, and I doubt anyone wants a shortcut for that. (They will be able to select all and delete when they need that.)

PR #1912, opened 2016-04-18
Crop and delete cell buttons are now enabled appropriately

shaneporter:
- Delete is only enabled when at least 1 cell is selected.
- Crop is only enabled when there is at least 1 selected cell **and** there is at least 1 non-selected cell. This is because crop deletes non-selected cells and we don't want it deleting all (when none are selected, as described in #1904.)


gordonwoodhull:
This is great. The keyboard command for crop needs to be enabled the same way, will open another issue.

Rebased and merging. Thanks @shaneporter!

PR #1913, opened 2016-04-18
Search resulting in no matches does not now cause an error.

shaneporter:
Wrapped a condition around the `match`, did some code indentation on the `switch` code block.


gordonwoodhull:
Yeah, I don't know why emacs wants to put `case` at the same level as `switch`, seems inconsistent.

gordonwoodhull:
Rebased and merged. thanks @shaneporter !

PR #1920, opened 2016-04-20
Crop cells shortcut does delete all if none selected

shaneporter:
Delegated selection bar code to new controller methods. 

Controller's `crop_cells` now uses `can_crop_cells` function.


gordonwoodhull:

PR #1921, opened 2016-04-20
CTRL (or CMD) click recent notebooks open in new tab

### LINKED ISSUE #1777: ctrl-click in recents menu should open in new tab
unfortunately we are special-casing control-clicks everywhere, to restore the browser's default behavior.

it would be helpful here.

shaneporter:
This resolves #1777. 

Caret introduced to 'Recent' text.


gordonwoodhull:
Thanks @shaneporter! Rebased and merged

Issue #1831, opened 2016-03-10
Esc <esc> key doesn't work properly for Editable titles and some dialog boxes


sanketd11:
- [x] The dialog box for import external notebook disappears after pressing Esc key. Similar behavior is not reproduced by group management dialog box. 
- [ ] Similarly for editable titles, the title field should go out of focus when Esc key is pressed. But currently after editing the notebook titles, if we press Esc key, it seems to be gone out of focus but the focus remains there.

gordonwoodhull:
Thanks @sanketd11. I remember there was some reason why Anatoliy didn't enable esc for the group management box. I can't remember it but let's enable it to see if is a problem.

shaneporter:
Not sure if there _was_ an issue here, perhaps fixed by e694e6f.

- The Group management dialog, like others, can be closed by pressing `esc`.
- Editable titles go out of focus when the `esc` key is pressed.

gordonwoodhull:
The dialog one is fixed. 

The editable title problem is still there. The first key typed after `esc` selects the title again, and then the second key starts replacing the title. I am not sure it is a huge problem, because where should the focus go? I guess we could set it explicitly to focus on the document.

gordonwoodhull:
I'm not sure if PR #1911 / e694e6f has any effect here.

gordonwoodhull:
Fixed by #1922

PR #1922, opened 2016-04-20
ESCaping after notebook title node edit removes focus (for real)

shaneporter:
This should fix the editing titles part of issue #1831. 

Setting focus on something else didn't work, so I used `window.getSelection().removeAllRanges()`.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -977,6 +977,13 @@ var editor = function () {
         var element = $li.find('.jqtree-element'),
             title = element.find('.jqtree-title');
         title.css('color', node.color);
+
+        title.on('keydown', function(e) {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Seems to work in notebook tree but not in navbar or comments. I don't think this issue is terribly important; probably I was on auto-pilot when I assigned this to the current release. I'll just close this and put the issue off.

gordonwoodhull:
Oh. It's probably because the fix should go into `ui_utils.editable` and not just editor_tab.

gordonwoodhull:
Put it there in above commit, tested, resolved! Thanks @shaneporter!

shaneporter:
Cool collaboration.

Issue #1840, opened 2016-03-14
Import button is disabled while importing Notebook from file


Prateek032:
1. Open import notebook from file 
2. select any one of the file 
3. Now press <ENTER> key, and again select a file
Import button is disabled
![import1](https://cloud.githubusercontent.com/assets/10941328/13741927/35d51154-e996-11e5-8660-1cc7f44c198a.PNG)

gordonwoodhull:
Thanks for reporting this, @Prateek032!

shaneporter:
I can only reproduce this if I select the _same_ file.

gordonwoodhull:
Ditto, I only see this for the same file. Presumably doesn't fire changed event if it's not changed. 😛 

Now testing the fix.

gordonwoodhull:
Duh, as you said in the PR. Thanks again.

PR #1924, opened 2016-04-21
Import Notebook from File - Import button is enabled appropriately

shaneporter:
This is a fix for issue #1840.

I gave the `file_select` event handler code a little more room to breathe, and added a couple of lines to the click handler.

If the user uploads the same file a successive time, the `change` handler is not invoked, so the 'Import' button, disabled by the `click` handler, was not being enabled as a result of `do_upload` success.

On click, the `file_select` value is set to null, meaning that any successful file upload, same or not, is interpreted as a change.

A caveat is that if the user clicks `cancel` on the upload, any previous file will have been cleared.


gordonwoodhull:

Issue #1897, opened 2016-04-11
Removing lines of code in notebook cell doesn't reduce cell's height


shaneporter:
Steps to reproduce:

1. Type enter/return in a cell to increase its number of rows to around 20.
2. Click the pencil icon above the cell to exit the cell's edit mode.
3. Click the pencil icon above the same cell.
4. Remove lines to observe that the lines are removed, but the height of the cell does not reduce.

![image](https://cloud.githubusercontent.com/assets/2493614/14431733/e8369b32-fffe-11e5-911e-066121ca1a83.png)

Note that the cell height does adjust correctly when the pencil is clicked again (exit cell's edit mode.)

Issue #1894, opened 2016-04-11
Complete code is not visible for the new cells


Prateek032:
When we write a code in new cell, the code content appears like below

![cell](https://cloud.githubusercontent.com/assets/10941328/14421203/e9a6f1ee-ffed-11e5-9c3e-85859e344b7f.PNG)

and its visible after executing cell or refreshing the tab

shaneporter:
The behaviour I see is that a new cell is created with 2 lines, but the cell's height increases automatically when I add a new line.

shaneporter:
This has been tackled by PR #1925.

PR #1925, opened 2016-04-21
Cell height expands/contracts appropriately

shaneporter:
#1894 
A new cell's height expands as expected as new lines are added.

#1897
This has also been fixed as a result of the change made for the above. As lines are removed, the cell's height contracts appropriately.


shaneporter:
closing for the moment - I've noticed some other weirdness.

Issue #1906, opened 2016-04-16
backspace in find dialog invokes cell deletion instead


gordonwoodhull:
likewise in notebook title, comments, settings, etc. (this is on mac of course, i assume there is similar behavior for del on windows.)

all of these are regular edit boxes or contenteditable - we can't steal delete or backspace from them. probably the code in init.js that disables backspace as browser-back but leaves it enabled for various edit controls, could be put in a function and reused here.

Prateek032:
Same behavior on Ubuntu

shaneporter:
this should now be fixed by commit e694e6f

sanketd11:
This issue is still reproducible for Find/Replace dialog only

sanketd11:
Confirmed ! This issue still persist on  dcdb296

shaneporter:
Have just checked PR #1933, and that fixes this issue too. @gordonwoodhull - please could you review and verify? Thanks.

gordonwoodhull:
Ah yes I didn't merge that because I was too confused. Will look at it with fresh eyes.

gordonwoodhull:
fixed by #1933

sanketd11:
This issue is still out there in ff2aec4 (latest commit)

gordonwoodhull:
Uh oh, I tested the heck out of this, what's going on.

shaneporter:
I have ff2aec4 and I'm unable to reproduce the issue.

Just so I'm not missing something, here's what I do on Ubuntu 14.04/Chrome.

1. Select one or more cells in a notebook.
2. Invoke find with `ctrl` `f`.
3. Type a search string into the 'find' input.
4. Press the backspace key.
5. A character is removed. Nothing else happens.

:confused:

shaneporter:
@gordonwoodhull, @sanketd11 - I have the latest develop branch code and I'm not able to reproduce this issue. Given that recent shortcut code has been merged, can this issue be closed? Thanks.

sanketd11:
Confirmed ! This issue is no longer reproducible.

Issue #1931, opened 2016-04-22
Typing ? in ACE editor invokes Help Dialog


shaneporter:
Typing a ? in any ACE editor (cells, prompt or asset) invokes the Help Dialog.

gordonwoodhull:
Whoops! Yeah, we probably need that key.

Prateek032:
Even for 'Find/Replace' invokes the help dialog

Issue #1928, opened 2016-04-22
esc to close Notebook Permissions / Group Management


Prateek032:
< esc > key  fails to close the modal window, when the focus is in any of the fields. In case of any other modal windows < esc > key closes the modal window even if the focus is on editable fields

gordonwoodhull:
Confirmed. Thanks @Prateek032.

PR #1933, opened 2016-04-26
? can now be typed where required, ESC closes dialogs when focus is in field

shaneporter:
By default, mousetrap prevents shortcuts being invoked in form fields. It runs the shortcut if the focus element has a class of `mousetrap`.  In a previous commit, I added the `ace_text-input` class to this check.

I have added an optional 'global' modifier to a shortcut so that it is always invoked regardless of focus element. This applies to the `esc` shortcut (#1928).

I have also modified the `window.Mousetrap.prototype.stopCallback` function so that it checks for 'text' entry (no `ctrl`, `meta` or `alt` modifiers) so currently, this applies to shortcuts invoked using `delete` and `backspace`. It only invokes the shortcut if it isn't text, and thus fixing #1931.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -197,7 +197,7 @@ RCloud.UI.init = function() {
             ['?']
         ],
         modes: ['writeable', 'readonly'],
-        action: function() { 
+        action: function(e) {
             if(!$('.modal').is(':visible')) {
                 RCloud.UI.shortcut_dialog.show(); 
             }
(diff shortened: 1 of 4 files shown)


shaneporter:
This PR should also fix #1906.

gordonwoodhull:
I reduced the double negatives a little bit and committed.

The design is certainly better than what we had, but the rules still feel kind of hokey to me. Some keys are global; we will allow shortcuts to work anywhere as long as they use a modifier key, what? But considering we had no structure at all before, it's good even to be able to ask these questions!

gordonwoodhull:
rebased and committed

shaneporter:
I agree. I'm not happy with the shortcuts stuff at the moment. It doesn't quite feel `right`.

I'm hoping there are no more functional issues, but I aim to make the code better. (Less `hokey`.)

PR #1934, opened 2016-04-26
Fixes for middle column ACE editor cursor disappearance and cell height

shaneporter:
So, I thought the `autoScrollEditorIntoView` property would help me out, but this only resulted in carnage. But I had to implement some custom handling for `setAutoScrollEditorIntoView` and pageup and pagedown handling to get things behaving as I'd like.

The situation in the middle column is complicated from a default ACE handling perspective, because the height of the middle column isn't the same as the height of an ACE editor, since there can be more than one editor in there, as well as all the other controls and status.

I started with a default implementation for `setAutoScrollEditorIntoView` and modified it for the needs of the middle column.

I also noticed that the page up/down behaviour wasn't ideal, since it shifted up/down the entire content length. This has also been changed.

I'm _hoping_ that these changes make the middle column a nicer place to be.


gordonwoodhull:
Now that is some adroit Ace hacking. Congratulations!

As far as I can tell, this makes the behavior exactly as if the whole notebook is one coherent editor, at least in this regard. Very nice. Of course, we'll have to put it in front of real users before we really know, but this feels like a huge improvement. It is such a relief to have the scroll-within-a-scroll gone!

gordonwoodhull:
Rebased and merged!

gordonwoodhull:
I was going to suggest that we abstract this and contribute it back to Ace, but it looks like they are deprecating the feature entirely. Hopefully it is getting replaced with something else, although at this point we could probably just do this externally anyway.

https://github.com/ajaxorg/ace/issues/2499

shaneporter:
I can't recall where I saw it, but there was talk about being able to specify the parent (with relation to scrolling.)

gordonwoodhull:
It was probably here, https://github.com/ajaxorg/ace/issues/2173#issuecomment-57917176, where the main maintainer nightwing says that Ace wasn't "optimized" for it and then comes up with a fix for it.

shaneporter:
Yep, that's it.

PR #1935, opened 2016-04-27
Discover page notebook title wrap fix for Firefox

shaneporter:
Introduced firefox-friendly CSS (it ignored the existing `word-break: break-word` property/value.)

### DIFF
--- a/htdocs/sass/rcloud-discover.scss
+++ b/htdocs/sass/rcloud-discover.scss
@@ -99,11 +99,8 @@ progress::-webkit-progress-value {
 .grid-item h1{
     font-size: 17px;
     margin: 0;
-    /* firefox */
     white-space: pre-wrap;
     word-break: break-all;
-    /* chrome */
-    word-break: break-word;
 }
 
 .grid-item h2 {
--- a/htdocs/sass/rcloud-discover.scss
+++ b/htdocs/sass/rcloud-discover.scss
@@ -99,6 +99,10 @@ progress::-webkit-progress-value {
 .grid-item h1{
     font-size: 17px;
     margin: 0;
+    /* firefox */
+    white-space: pre-wrap;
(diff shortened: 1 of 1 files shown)


htdocs/sass/rcloud-discover.scss
```diff
@@ -99,6 +99,10 @@ progress::-webkit-progress-value {
 .grid-item h1{
     font-size: 17px;
     margin: 0;
+    /* firefox */
+    white-space: pre-wrap;
+    word-break: break-all;
+    /* chrome */
     word-break: break-word;
```
gordonwoodhull: Maybe the issue is that this value doesn't exist? It's a `word-wrap` value instead. Confusing.

shaneporter: Chrome is happy with it, but it's considered an invalid value for the `word-break` property by Firefox.

gordonwoodhull: Yeah, it's [not on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), so I'm guessing it's non-standard. Chrome seems to be happy with `break-all`, could you verify and update your PR?

shaneporter: I've checked the visual output in Chrome and Firefox on Ubuntu, and it is correct. Apologies if I've misinterpreted your _verify and update_ statement.

gordonwoodhull: Sorry, I was unclear! I meant that I think there's no need to special case Chrome with something non-standard when the standard `word-break: break-all` seems to work on both browsers.

shaneporter: Haha, no problem.

gordonwoodhull:
Thanks @shaneporter!

Issue #1914, opened 2016-04-18
make find/replace dialog less ugly


gordonwoodhull:
I was in a rush and basically didn't style it at all, beyond functionally. 

Its looks do not inspire confidence (I guess they match the bugginess in the code).

shaneporter:
Submitted PR #1936.

PR #1936, opened 2016-04-27
Beautification of find/replace dialog

shaneporter:
This attempts to address #1914.

I have implemented an underscore template in `edit.html` (0ec5c12) which makes working with the markup a little easier, and removes some boilerplate DOM manipulation from `find_replace.js`.

Other than that, it's pretty much visual styling. Hopefully it's prettier.

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -63,6 +63,30 @@
   </span>
 </script>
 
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Much improved in look and code. Thanks @shaneporter!

Rebased and merged.

PR #1946, opened 2016-04-28
Shortcuts fixed (including CTRL + S) for 'Discover'

shaneporter:
There are currently no applicable shortcuts for the Discover page, so the relevant line has been removed.

The impact is that all browser defaults are active.


gordonwoodhull:
Fair enough. May need a more general solution in the future, but I have no idea what kind of shortcuts this page would need anyway.

gordonwoodhull:

PR #1961, opened 2016-05-05
Removed mini.html part from the SmokeTest.js

Prateek032:


gordonwoodhull:
Awesome, it looks like there are a ton of other fixes in here too. Let's see how it goes!

gordonwoodhull:
Success with this commit! I'll push a bunch of Shane's fixes today and keep an eye on it.

Issue #1929, opened 2016-04-22
Firefox: Cells getting selected while moving cells 


Prateek032:
While moving cells from top to bottom and bottom to top, the cell is getting selected in Firefox browser

gordonwoodhull:
Apparently it registers a click after the event is done.

To continue the file browser analogy, this is like having some files selected and then starting a drag on another file that is not selected. It looks like the OSes disagree here: Windows and Linux will select just the dragged file and deselect the the old ones, while OSX will drag the other file while keeping the old selection.

I slightly prefer the Windows/Linux approach, in which case we should change the selection on mousedown. But this is open for debate - alternately we could ignore the click-after-drag on Firefox.

Somewhere I started some notes on what multisortable behavior would mean for notebook cells. Not something we should worry about for this release!

gordonwoodhull:
Fixed in 956b801906df3e1e2f5f925bebdae9da494fc684 (merged in #1962)

PR #1962, opened 2016-05-05
Shortcuts are not invoked whilst a dialog is visible

### LINKED ISSUE #1910: Unable remove admin/member from group
Steps to replicate:
- Open manage groups dialog box
- Press 'Delete' key
- Try to remove any Member or Admin from group

After pressing 'Delete' key the cursor will shift to command prompt which doesn't come back to current dialog box

shaneporter:
I have added an `enable_in_dialogs` property to shortcuts. By default this is false and means that when a dialog is shown, an action does not occur. This fixes #1910.

I have modified the 'esc' -> close dialog shortcut so that this property value is true. Because we want to be able to close a dialog.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -23,6 +23,7 @@ RCloud.UI.init = function() {
                 shell.notebook.controller.move_cell(model, next);
             },
             handle: " .cell-status",
+            helper: 'clone',
             scroll: true,
             scrollSensitivity: 40,
             forcePlaceholderSize: true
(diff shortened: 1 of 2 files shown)


htdocs/js/ui/init.js
```diff
@@ -23,6 +23,7 @@ RCloud.UI.init = function() {
                 shell.notebook.controller.move_cell(model, next);
             },
             handle: " .cell-status",
+            helper: 'clone',
```
gordonwoodhull: I think this snuck in here from other work, committing it anyway.

shaneporter: Sorry! That was the fix for #1929.

gordonwoodhull:
Rebased and merged. Thanks @shaneporter, looks like a better solution!

PR #1963, opened 2016-05-05
Resolves Firefox cell select after move bug

shaneporter:
The simple change resolves the issue and Chrome remains happy.


gordonwoodhull:
Fix was included in #1962

Issue #1899, opened 2016-04-13
Undo feature is not working for 'Read only' notebooks


Prateek032:
Open other user Notebook and press < Ctrl + Z > (undo), previous History version of the notebook is not opening

gordonwoodhull:
Confirmed.

gordonwoodhull:
It is kind of weird to undo a readonly notebook, but since we are already abusing a well-known feature, we should be consistent about it.

Prateek032:
This issue still persist on latest bae8152 commit

gordonwoodhull:
So #1976 / #1971 interferes here. Looks like undo and redo commands should be available in edit mode but not view mode. It's not about the notebook status, just the page that is open, for this one.

gordonwoodhull:
Thanks @Prateek032!

gordonwoodhull:
Noticed this is still not working, looked back at the history. Look at these consecutive commits on init.js:

```
commit 1175ada94d5176ca8150ada118ae75ebd2deaab7
Author: shane <shane@freshclickmedia.com>
Date:   Fri May 13 11:35:45 2016 +0100

    fixes #1971 - undo and redo are not available in view mode

diff --git a/htdocs/js/ui/init.js b/htdocs/js/ui/init.js
index 98d9437..fe5303b 100644
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -141,6 +141,7 @@ RCloud.UI.init = function() {
             ['command', 'z'],
             ['ctrl', 'z']
         ],
+        modes: ['writeable'],
         action: function() { editor.step_history_undo(); }
     }, {
         category: 'Notebook Management',
@@ -150,6 +151,7 @@ RCloud.UI.init = function() {
             ['ctrl', 'y'],
             ['command', 'shift', 'z']
         ],
+        modes: ['writeable'],        
         action: function() { editor.step_history_redo(); }
     }]);
 

commit 2c6fdbcc6947a15e9bebb0b2b72451a6ad38293e
Author: shane <shane@freshclickmedia.com>
Date:   Thu May 5 16:33:28 2016 +0100

    fixes #1899, history step forwards also enabled

diff --git a/htdocs/js/ui/init.js b/htdocs/js/ui/init.js
index 8276476..98d9437 100644
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -141,7 +141,6 @@ RCloud.UI.init = function() {
[...]

PR #1965, opened 2016-05-06
Undo/redo issues resolved

### LINKED ISSUE #1876: undo/redo (Ctrl+Z / Y) operation fails for second attempt
Steps to replicate:
- Load any notebook having several history versions 
- Press Ctrl+ Z ( This will load the history version of that notebook)
- Now press Ctrl +Z again ( This time nothing will happen )

For the next undo operation to work user needs to click on screen. Same is the behavior for 'redo' operation.

shaneporter:
This pull request is aimed at fixing #1876 and #1899.

#1899 was a simple fix, removing the `modes` property value of `writeable` from the undo/redo shortcuts, so they use the default of 'writeable' and 'readonly'. 

#1876 required a little more work, and I noticed that the shortcut code's use of the `modes` property was only happening at the point they were added. I modified the code so that the shortcuts are added, and when they are invoked, it checks to see if they should be actioned. 

To fix #1876, I took an approach that when the user steps back (undo) or forwards (redo) in the notebook history using a shortcut, that they should have to wait until that notebook is loaded before stepping again. I tried repeatedly undoing, and from the user's perspective, it's not easy to know exactly what's happening.

To that end, I introduced the `disable` and `enable` functions to the `shortcut_manager`. They both take an array (or single string) for the IDs of the shortcuts that should be disabled or enabled. When the history manager's `step_history_undo` and `step_history_redo` functions are called, the undo/redo shortcuts are disabled. When the notebook has loaded, the shortcuts are enabled.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -1476,17 +1476,27 @@ var editor = function () {
                 });
         },
         step_history_undo: function() {
+
+            RCloud.UI.shortcut_manager.disable(['history_undo', 'history_redo']);
+
             var previous_version = history_manager.get_previous();
 
             if(!_.isUndefined(previous_version)) {
-                this.load_notebook(current_.notebook, previous_version);
+                this.load_notebook(current_.notebook, previous_version).then(function() {
+                    RCloud.UI.shortcut_manager.enable(['history_undo', 'history_redo']);
+                });
             }
         },
         step_history_redo: function() {
+
+            RCloud.UI.shortcut_manager.disable(['history_undo', 'history_redo']);
+
             var next_version = history_manager.get_next();
 
             if(!_.isUndefined(next_version)) {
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Okay, this seems like a decent workaround for now. 

In fact, now that I think about it, we might use a solution like this to disable the user interface while in the process of loading a different notebook / version. It's kind of accidental that commands can get through while we're in that process but it has to do with our extensive use of asynchronous programming. (Executing another command while loading a notebook seems like something that can't be made to work.)

In any case, it's easy to type key commands really fast, so this is the most important place for this. We may find other keys that also need to be disabled while notebooks are loading (and we may need to move the enable/disable logic deeper into notebook loading). But this is a good start.

gordonwoodhull:

Issue #1819, opened 2016-03-09
selection bar still takes up space when hidden


gordonwoodhull:
When the notebook is readonly (someone else's or a historical version), the selection bar is hidden, as it should be. Yet the notebook height still takes it into consideration, so there's a gap at the bottom of the screen.

![image](https://cloud.githubusercontent.com/assets/1366709/13641866/2c58d6a8-e5e8-11e5-85b3-bd071369f960.png)

gordonwoodhull:
Related (not that you have to think about it @shaneporter, but maybe there's a common solution) #1723 search/replace bar takes up too much space. Maybe there is a little vertical space manager we need here.

gordonwoodhull:
fixed by #1829

PR #1966, opened 2016-05-06
This fixes the gap in the middle column (when selection bar isn't shown) issue

shaneporter:
So, this happened in the past (#1819), but reappeared, perhaps something to do with changes made for the fix for #1934), but I'm not sure.

A small change has resolved the issue.


gordonwoodhull:
@shaneporter, thanks for catching this regression and fixing it. Looks like the (s)css is a tad cleaner this way too.

Rebased and merged.

PR #1967, opened 2016-05-06
This fixes range selection in notebook cells

shaneporter:
I introduced a `last_selected` variable to keep track of the last selected cell's index. This is updated when the user clicks on the header of a cell (exclusive) or clicks on a cell's selection checkbox (additive).

The range selection uses this value to determine which cells to check following a range (invoked with shift) operation is undertaken.

### DIFF
--- a/htdocs/js/notebook/notebook_model.js
+++ b/htdocs/js/notebook/notebook_model.js
@@ -2,8 +2,9 @@
 /*jshint -W083 */
 Notebook.create_model = function()
 {
-    var readonly_ = false;
-    var user_ = "";
+    var readonly_ = false,
+        user_ = "";
+        last_selected_ = undefined;
 
     function last_id(cells) {
         if(cells.length)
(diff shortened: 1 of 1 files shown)


htdocs/js/notebook/notebook_model.js
```diff
@@ -2,8 +2,9 @@
 /*jshint -W083 */
 Notebook.create_model = function()
 {
-    var readonly_ = false;
-    var user_ = "";
+    var readonly_ = false,
+        user_ = "";
+        last_selected_ = undefined;
```
gordonwoodhull: Here's an accidental global variable. (Damn you, JavaScript!)

shaneporter: Yeah, you're right. Sorry about that.

gordonwoodhull:
I suspect there are more subtleties of shift-select that we are missing, but this is very close.

gordonwoodhull:

Issue #1958, opened 2016-05-05
Replace field appears in history version of the notebook


Prateek032:
steps to replicate:
1. Open any notebook
2. Invoke Find/Replace fields
3. Now click on History icon and open any History version of that notebook or simply press < ctrl + z >

![image](https://cloud.githubusercontent.com/assets/10941328/15036758/2476d5fc-12b1-11e6-8fba-4ea2a8113ff7.png)

gordonwoodhull:
Confirmed. I think we enable the keyboard shortcuts based on the readonly mode but we don't similarly modify the find/replace dialog.

For better or worse, there's a whole source file dedicated to enabling/disabling things based on the readonly mode.

`htdocs/js/ui/configure_readonly.js`

gordonwoodhull:
Thanks @Prateek032.

PR #1968, opened 2016-05-09
Replace fields hidden for read-only notebooks

shaneporter:
I considered there to be two issues with #1958. The first was that the 'replace' fields could be shown for read-only notebooks using the keyboard shortcut. That was implemented with the fix for #1876 (submitted with PR #1965).

The second issue is that the 'replace' fields are not hidden when (using undo) the notebook changes from writable to read-only. Commit 5fe27bd fixes this with two straightforward modifications, described in the changed files below.

### DIFF
--- a/htdocs/js/ui/configure_readonly.js
+++ b/htdocs/js/ui/configure_readonly.js
@@ -28,6 +28,7 @@ RCloud.UI.configure_readonly = function() {
             .prop('checked', false)
             .attr("disabled", true);
         RCloud.UI.scratchpad.set_readonly(true);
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
This borks if you load a readonly notebook without having previously opened the find/replace dialog. (E.g. if the current notebook/version is readonly and the page is reloaded.)

shaneporter:
I'll take a look at that!

shaneporter:
Able to reproduce, and hopefully fixed with that last commit.

gordonwoodhull:
Thanks @shaneporter it's all happy now.

Rebased and merged.

Issue #1960, opened 2016-05-05
Cell contents are replaced in another notebook


Prateek032:
Steps to replicate:
1. Open a notebook
2. Invoke Find/Replace and replace the contents with the existing content
3. After replacing switch to another notebook
4. Now again click on replace.
The cell contents are getting replaced

![replace](https://cloud.githubusercontent.com/assets/10941328/15039110/a6e76cd4-12c7-11e6-9ada-bb531424a33e.gif)

gordonwoodhull:
As we discussed on the call, this is probably another aspect of #1849.

Thanks @Prateek032! BTW, this is helpful but there is no need to continue testing find&replace until we fix these issues - it is clear that it is not tracking any changes to the notebook (or indeed, the notebook being replaced entirely!)

shaneporter:
...and #1839.

PR #1969, opened 2016-05-10
Find/replace bugs suggested fix

shaneporter:
So, I'm not sure that this is ready to merge as it stands, but here's the pull request for you to look at.

I've added some code to the `on_dirty` function of `notebook_controller` to refresh the search results. This uses underscore's `debounce` function, with a `wait` parameter of 350ms, to rate limit the call of `find_replace`'s `reset` function. This redoes the searching.

Before these changes, the highlighting can get really screwed up, and this can occur occasionally because of that 350ms `wait` value.

I feel that the changes for #1960 (1017dde) are less contentious than the other fixes. `load_notebook` now resets the search.

Feedback welcome. Thanks.

### DIFF
--- a/htdocs/js/notebook/notebook_controller.js
+++ b/htdocs/js/notebook/notebook_controller.js
@@ -26,6 +26,10 @@ Notebook.create_controller = function(model)
         };
     }();
 
+    var refresh_search = _.debounce(function() {
+        update_notebook(refresh_buffers()).then(function() { RCloud.UI.find_replace.reset(); })
+    }, 350);
+
     function append_cell_helper(content, type, id) {
         var cell_model = Notebook.Cell.create_model(content, type);
         var cell_controller = Notebook.Cell.create_controller(cell_model);
@@ -228,6 +232,8 @@ Notebook.create_controller = function(model)
                 result.save();
                 save_timer_ = null;
             }, save_timeout);
+
+        refresh_search();
     }
 
     model.dishers.push({on_dirty: on_dirty});
--- a/htdocs/js/ui/find_replace.js
+++ b/htdocs/js/ui/find_replace.js
@@ -295,6 +295,13 @@ RCloud.UI.find_replace = (function() {
             if(replace_stuff_) {
                 replace_stuff_.hide();
             }
+        },
+        reset: function() {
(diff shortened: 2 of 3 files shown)


htdocs/js/ui/find_replace.js
```diff
@@ -295,6 +295,20 @@ RCloud.UI.find_replace = (function() {
...
+                active_match_ = undefined;
+                build_regex(find_input_.val());
+                highlight_all();
+            }
+
+        },
+        update_search: function() {
+            if(find_dialog_ && find_dialog_.is(':visible')) {
+                this.reset();
+            }
```
gordonwoodhull: New function. Where does it get called?

shaneporter: Looks like I committed the `update_search` function erroneously. It was used by `notebook_controller` at a point, but not at the point I committed it.

It is superfluous. Please ignore it.

htdocs/js/notebook/notebook_controller.js
```diff
@@ -26,6 +26,10 @@ Notebook.create_controller = function(model)
         };
     }();
 
+    var refresh_search = _.debounce(function() {
+        update_notebook(refresh_buffers()).then(function() { RCloud.UI.find_replace.reset(); })
+    }, 350);
+
```
gordonwoodhull: This is forcing a save to GitHub which seems like overkill (and disablesj/falsifies the autosave feature that is the rest of `on_dirty`). 

Is this because find/replace will only look at the notebook model content rather than the current live content of the ace editors?

gordonwoodhull:
It seems like most editors simply disable search results unless focus is on the find or replace input. Thus practically you can't change the notebook while search results are active. We could consider that if the dichotomy between ace buffer and notebook model makes the approach in this PR too difficult.

gordonwoodhull:
@shaneporter, should we close this one since we're doing a different approach?

shaneporter:
Yes, the other approach is significantly different. Thanks.

gordonwoodhull:
Okay, closing this. Trying to keep search in sync while editing the notebook seems like too much of a hassle, especially if it means lots of extra commits.

PR #1974, opened 2016-05-13
Asset and Group names are trimmed prior to name check

shaneporter:
`trim()` to the rescue.

### DIFF
--- a/htdocs/js/notebook/asset_view.js
+++ b/htdocs/js/notebook/asset_view.js
@@ -24,18 +24,21 @@ Notebook.Asset.create_html_view = function(asset_model)
                 filename_span.text(old_asset_name);
                 return;
             }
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
We'll need this in at least two more places: Group Rename, and Asset Rename. I'll just leave this open so you don't have the overhead of opening another PR.

shaneporter:
dd48788 should fix the group rename.

I started looking at asset rename and found another issue. If you perform a rename with the _same_ name, it complains that an asset of that name already exists.

shaneporter:
Hi @gordonwoodhull - e7aa133 now checks if the user has changed the asset name before checking if it exists. Please re-review this PR. Thanks.

gordonwoodhull:

PR #1975, opened 2016-05-13
Dragging cell retains cell language select value

shaneporter:
jQuery's clone doesn't retain `select` values. It's a known issue, [described here](http://stackoverflow.com/questions/742810/clone-isnt-cloning-select-values).

Of particular interest is [this answer](http://stackoverflow.com/questions/742810/clone-isnt-cloning-select-values#answer-11804162).

I utilised the suggested jquery plugin, `jquery.fix.clone`, and it seems to be working. I tested in Firefox and Chrome.

Issue #1929 has not regressed has a result of this change.


gordonwoodhull:
Nice. Rebased and merged. Thanks @shaneporter!

PR #1979, opened 2016-05-18
Added 'show keyboard shortcuts' to help pane.

shaneporter:
Not sure if you'll like the keyboard (`<i class="icon-keyboard"></i>`) in the link, but I won't be offended if you remove it.

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -133,7 +133,7 @@
 <script type="text/html" id="help-snippet">
   <div id="help-wrapper" class="panel-body widget-vsize">
     <a href="#" id="show-shortcuts">
-      <i class="icon-keyboard"></i>Keyboard shortcuts...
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
heheh, cute.

gordonwoodhull:
This is great. Only change I'd like to see is that, since vertical real estate is precious, this should fit to the right of the search form. I tried to fix it myself and got bogged down in CSS/HTML, so I'm going to pass this back to you. 

We can trim the text down to just "keyboard shortcuts" or even "keyboard", if it won't fit otherwise.

shaneporter:
I changed the link text to 'Keyboard', but kept the little keyboard icon.

At a smaller window size, it is possible to overlap the link with the help input/button, but I don't know how many users will encounter this issue during normal use.

Please re-review @gordonwoodhull. Thanks. :smile:

gordonwoodhull:

PR #1980, opened 2016-05-18
'Revert' keyboard shortcut

shaneporter:
Went for CTRL/CMD + e.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -159,7 +159,6 @@ RCloud.UI.init = function() {
             ['ctrl', 'e'],
             ['command', 'e']
         ],
-        modes: ['writeable'],
         action: function() { 
             editor.revert_notebook(shell.notebook.controller.is_mine(), shell.gistname(), shell.version());
         }
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -160,7 +160,7 @@ RCloud.UI.init = function() {
             ['command', 'e']
         ],
         action: function() { 
-            if(shell.notebook.controller.is_mine()) {
+            if(shell.notebook.controller.is_mine() && !shell.is_view_mode()) {
                 editor.revert_notebook(shell.notebook.controller.is_mine(), shell.gistname(), shell.version());
             }
         }
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -159,6 +159,7 @@ RCloud.UI.init = function() {
             ['ctrl', 'e'],
             ['command', 'e']
         ],
+        modes: ['writeable'],
         action: function() { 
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Nope, if you prevent revert when the notebook is not writeable, then you prevent it entirely. 😛 

Confirmed this works before the last commit.

shaneporter:
Yes, 04b4f9e wasn't my smartest ever commit. :disappointed: 

I was thinking about other people's notebooks, but the best way was to enable revert when not writeable and check with `is_mine()`.

Please re-review this PR. Thanks.

gordonwoodhull:
This is great. So nice to be able to undo and then revert with just keys.

This does open up one new bug, which is that revert is now available from view mode. So if you're viewing a historical version, you can invoke revert, which reverts but then throws an error:

```
TypeError: Cannot read property 'set_url' of undefined
    at set_page (http://127.0.0.1:8080/js/rcloud_bundle.js:9650:55)
From previous event:
    at Object.update_link (http://127.0.0.1:8080/js/rcloud_bundle.js:9717:18)
    at http://127.0.0.1:8080/editor_tab.js:1630:59
From previous event:
    at http://127.0.0.1:8080/editor_tab.js:1628:39
From previous event:
    at Object.revert_notebook (http://127.0.0.1:8080/editor_tab.js:1453:18)
    at Object.action (http://127.0.0.1:8080/js/rcloud_bundle.js:7089:24)
    at func_to_bind (http://127.0.0.1:8080/js/rcloud_bundle.js:6247:54)
    at g (http://127.0.0.1:8080/lib/js/mousetrap.min.js:5:35)
    at c.d._handleKey (http://127.0.0.1:8080/lib/js/mousetrap.min.js:7:211)
    at c.handleKey (http://127.0.0.1:8080/lib/js/mousetrap.min.js:10:272)
    at HTMLDocument.e (http://127.0.0.1:8080/lib/js/mousetrap.min.js:5:260)
```

shaneporter:
You also get an error in view mode when invoking 'revert' for the current version (noted for testing)

```
Error: attempted to revert current version
    at Object.result.revert_notebook (editor_tab.js:1451)
    at Object.RCloud.UI.shortcut_manager.add.action (init.js:164)
    at func_to_bind (shortcut_manager.js:125)
    at g (mousetrap.min.js:5)
    at c.d._handleKey (mousetrap.min.js:7)
    at c.handleKey (mousetrap.min.js:10)
    at HTMLDocument.e (mousetrap.min.js:5)
```

shaneporter:
@gordonwoodhull - this has now been resolved with the addition of `!shell.is_view_mode()` in the condition for the shortcut. (0ae28e7)

gordonwoodhull:
Okay, this is somewhat misleading because the shortcut still appears in help. We might consider some kind of exclusive modes so that we can just say edit or not edit in the `modes` list. But this is good enough for now.

Rebased and merged.

PR #1982, opened 2016-05-19
Panel order change, setting to revert

### LINKED ISSUE #1802: Move "Help" and "Search" to the right pane
I have seen too many users constantly resizing the Notebook pane because of the Help/Search panes blowing up the Notebook pane and this obscuring the notebook itself. We should move the Help/Search panes to the right side where they belong and are compatible with the wide size of the Assets.

(Yes, I'm pretty sure this is a duplicate but can't find it and I'd really like to keep this very specific so it gets done :))

shaneporter:
Modified the panels so that by default, they are in a different order (associated by size.)

Added a checkbox for the 'Arrange Panels by Size' setting, defaulting to 'true'. Changing requires a reload.

On load, the setting is read and panels are rearranged if necessary.

This fixes #1802.

### DIFF
--- a/htdocs/js/ui/panel_loader.js
+++ b/htdocs/js/ui/panel_loader.js
@@ -1,5 +1,6 @@
 RCloud.UI.panel_loader = (function() {
     var extension_;
+    var panel_data_ = {};
     var panels_ = {};
 
     function collapse_name(name) {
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Reverting to the old layout doesn't work for me:

image

I guess the logic is backward (it's enforcing the new layout when it sees the option disabled, instead of restoring the old layout). Will fix.

gordonwoodhull:

PR #1983, opened 2016-05-19
Enable/disable extensions text supports comma, slash and space separators

shaneporter:
Regex mod; `filter` has been removed because I don't think it's required.

### DIFF
--- a/htdocs/js/ui/settings_frame.js
+++ b/htdocs/js/ui/settings_frame.js
@@ -132,7 +132,7 @@ RCloud.UI.settings_frame = (function() {
         text_input_vector: function(opts) {
             opts = _.extend({
                 parse: function(val) {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @shaneporter! You're right, it's cleaner to just capture all the delimiters with a star and not have to worry about empty items. 

What's the rationale for slash? I think that's pretty rare as a list delimiter.

shaneporter:
Don't know really... guess I was feeling in a 'slash' kinda mood.

gordonwoodhull:
Heheh, okay. It's interesting how my english language parser was paged out, I didn't even think of that. 

As you say, `/` means "or" not "and", so I'm going to remove it.

gordonwoodhull:
Drat, this still allows an empty primary or final empty item. So I think it still needs the filter. I'll close this and just change the regex so accepts (and corrects) space.

Issue #1987, opened 2016-05-20
Keyboard shortcut help content should be consistent across the browsers


Prateek032:
Currently when we invoke it in Firefox browser it looks like below

![image](https://cloud.githubusercontent.com/assets/10941328/15417470/dddff4d8-1e73-11e6-8edb-b11e8bafc3e1.png)

and in Chrome 

![image](https://cloud.githubusercontent.com/assets/10941328/15417493/2294ff7e-1e74-11e6-891b-268256d369c8.png)

shaneporter:
Resolved in PR #1990.

Issue #1949, opened 2016-04-28
run notebook from here on


gordonwoodhull:
adopting the opinion that "shift" sometimes means "moreso", i suggest the following shortcuts for running a cell and all cells following it

 - [x] shift-click play button on the cell
 - [x] shift-alt-enter within the cell

open to debate! please cite precedents if any.

shaneporter:
Implemented in PR #1990

Issue #1439, opened 2015-05-12
keyboard shortcut for creating new cell and navigating between cells


holyguster:
I was asked if this was something on the roadmap.  Figured I'd open an issue to discuss.

The specific ask was if there is a shortcut key combo for (1) creating a new cell, and (2) moving cursor focus to next/previous cell.

For (1), things to consider:
 - where would the new cell be created?  at the end of the notebook, below the current cell, above the current cell?  Maybe different key combo for each, or does that get to complicated?
 - what type of cell would be created?  use the type setting for the terminal cell, or use the type setting for the current cell?

For (2) and (3), I would imagine that the UI would treat the same as if the user clicked on the cell; if not already in edit mode, navigating to a cell would cause the cell to go into edit mode.

shyamram:
I was looking for keyboard short cut to execute a cell. Is there one? If not, can that be added?

gordonwoodhull:
To execute the current cell, press alt-enter. To create a new cell from the current selection or live, press ctrl-enter (or cmd-enter).

shyamram:
thank you.

gordonwoodhull:
I am in favor of this feature, because the keyboard is faster (and not having to switch between mouse and keyboard is even faster).

I want to avoid religious wars about what the keys should be, so if @s-u knows of equivalents in the R world, we should use them. But I think it's kind of wide open since the notebook/cell interface is pretty new.

IPython uses emacs-like ctrl-m prefixes, [so that's right out](http://ipython.org/ipython-doc/1/interactive/notebook.html#keyboard-shortcuts). But it's interesting to see what commands they have found useful.

I think there should be goto-last,goto-next, insert-after, insert-before, and goto-prompt (since that already inserts at end). I also think there should be delete-last, for folks using the command prompt and getting frustrated by all the cells being created. I don't yet have an opinion what the keys should be.

gordonwoodhull:
Not getting to this feature for 1.4. It isn't hard but it needs to be spec'd out because there are potentially very many keys.

shaneporter:
I was using IPython today, and noticed that:

- keydown on last row of current cell focuses on line 1 of next cell (or prompt);
- keyup on first row of current cell focuses on last line of previous cell.

gordonwoodhull:
Hmm, that may work. Of course right now we use the same behavior on the prompt for command history, so it won't be symmetric. I wonder what others think of this idea.

Thanks for calling attention to this issue, it's a vital part of #300.

[...]

PR #1990, opened 2016-05-23
Cell/shortcut enhancements

shaneporter:
This PR addresses 2 enhancements: #1439 and #1949. It also fixes the shortcut dialog ordering inconsistency #1987.

### #1439 ###

I haven't yet been able to implement an arrow left/right into previous/next cell shortcut. If you add an action to an existing shortcut, that existing shortcut is overridden. 

Alt-up, alt-down don't actually go to the start of the previous/next cell. I reused some existing code, and the cursor will reuse an existing cursor point, if it exists. If not, it will position the cursor at the top of the cell. I realise this is different to the enhancement's requirement, but I think it's an interesting variation that might be a better solution. 

I have implemented ctrl-[ and ctrl-] for insert cell before/after current.

These shortcuts are implemented using ACE, but have been documented using the shortcut manager, so that they appear in the shortcuts dialog.

### #1949 ###

Implemented as a shortcut and shift-click play icon on individual cell.

---

### #1987 ###

Quick fix for this one to ensure consistent ordering between Firefox/Chrome.

### DIFF
--- a/htdocs/js/notebook/cell_model.js
+++ b/htdocs/js/notebook/cell_model.js
@@ -34,6 +34,7 @@ Notebook.Cell.create_model = function(content, language)
         set_focus: function() {
             this.notify_views(function(view) {
                 view.edit_source(true);
+                view.scroll_into_view(true);
             });
         },
         deselect_cell: function() {
--- a/htdocs/js/notebook/cell_model.js
+++ b/htdocs/js/notebook/cell_model.js
@@ -31,6 +31,11 @@ Notebook.Cell.create_model = function(content, language)
                 version: this.parent_model.controller.current_gist().history[0].version
             };
         },
+        set_focus: function() {
+            this.notify_views(function(view) {
+                view.edit_source(true);
+            });
+        },
         deselect_cell: function() {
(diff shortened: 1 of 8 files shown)


gordonwoodhull:
@shaneporter, I agree that using the current position (if any) makes a lot more sense than going to the beginning of the cell. Good call.

As for arrowing to the next/previous cell, what we discussed in #1439 is using the up and down arrow keys. It's true that you can't just grab the keys; you'll need to "override" them by first fetching the current function for those keys, then install your own and call into the old function when appropriate. It's exactly the same logic we use in the command prompt for history, so you should be able to steal or reuse that.

I'll open a new issue for that, since this is ready to merge.

gordonwoodhull:
I'm also changing the insert cell keys to cmd-[ and cmd-] on mac.

gordonwoodhull:
Rebased and merged. It was a messy merge due to it stomping on the same lines as a lot of recent PRs, so I did it in a slightly different order.

gordonwoodhull:
Thanks @shaneporter!

Issue #1791, opened 2016-01-28
blurring edit mode on cells


gordonwoodhull:
We will need to at least "blur" (unfocus) any cell or the command prompt when a cell selection event happens for #658, so that we take keyboard commands directly.

Currently we don't deactivate cells unless they are run or you click the edit button manually. The rationale is that we don't want to fully deactivate because it can cause cell height changes and disorientation. 

So I guess I've just convinced myself that we still don't to deactivate cells automatically, but we do need to blur them.

Anatoliy suggested that there should be a manual key to deactivate when you are editing a cell, probably ESC. I think this makes sense.

gordonwoodhull:
@sanketd11, @Prateek032, @shaneporter, we discussed this today.. here's the issue I filed in January. (Makes you wonder what other gems might be buried in the tracker.)

Since the description above is kind of rambly, the two points are:
 - [ ] selecting or deselecting any cell should blur any active cell (or prompt or asset)
 - [ ] ESC key inside the active cell should blur it

Even though deactivation is now theoretically stable, I don't want to go that far yet. There doesn't seem to be any expense in keeping Ace active (but there is a big expense in activating the first time, which is why we do this).

shaneporter:
If a cell is active, it is implied to be in the editable state.

You're just talking about blurring its focus on `ESC`, rather than setting it to read only (as if the user clicked the 'edit' button for that cell. Is that right? 

Is deactivation the same as setting it to read only?

Thanks.

gordonwoodhull:
I'm just talking about the keyboard focus here. Probably my first description was too confusing and I should remove it.

"Read only" for me means that there is no way to edit it, like someone else's notebook. "Activated" means that the Ace editor is live, but not necessarily having the focus, whereas "deactivated" means that it is our html imitation of the Ace look (and a slightly lighter background).

shaneporter:
thanks for the clarification.

shaneporter:
A fix has been suggested with PR #1991.

gordonwoodhull:
Clarified title.

PR #1977, opened 2016-05-16
Delete cell shortcut works following cell selection by header click

shaneporter:
A suggested fix for #1930.

PR #1991, opened 2016-05-24
ESC or select cell to blur ACE

shaneporter:
This PR contains a number of changes for the implementation of #1791.

ACE widget is blurred as a result of:

- Selection of cell by checkbox or header
- Pressing `esc` from inside a cell.

### DIFF
--- a/htdocs/js/notebook/cell_model.js
+++ b/htdocs/js/notebook/cell_model.js
@@ -45,9 +45,15 @@ Notebook.Cell.create_model = function(content, language)
 
             this.notify_views(function(view) {
                 view.selected_updated();
+                view.deactivate_cell();
             });
 
             return is_selected_;  
+        },
+        deactivate_cell: function() {
+            this.notify_views(function(view) {
+                view.deactivate_cell();
+            });
         },
         toggle_cell: function() {
             is_selected_ = !is_selected_;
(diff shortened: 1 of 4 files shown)


htdocs/js/notebook/notebook_model.js
```diff
@@ -303,6 +303,10 @@ Notebook.create_model = function()
                 select_range(Math.min(start, end), Math.max(start, end));
             }
 
+            _.each(this.cells, function(cell) {
+                cell.deactivate_cell();
+            });
+
```
gordonwoodhull: #1977 already covered blurring of cells when the selection changes, so this is unnecessary, removing it.

htdocs/js/notebook/cell_model.js
```diff
@@ -45,10 +45,16 @@ Notebook.Cell.create_model = function(content, language)
 
             this.notify_views(function(view) {
                 view.selected_updated();
+                view.deactivate_cell();
```
gordonwoodhull: see note below, removing this in 81dceb017

htdocs/js/notebook/notebook_model.js
gordonwoodhull: on second thought, reverted 42695cf428b4b9470 in favor of this, it's more explicit what's going on

htdocs/js/notebook/cell_model.js
gordonwoodhull: as below, reverted the other fix instead

gordonwoodhull:
This superceded #1977, so I reverted that in favor of this. That one was more clever but this was more clear.

gordonwoodhull:

gordonwoodhull:
this caused #1999

PR #1995, opened 2016-05-25
Clearer error reporting in mini and shiny.

shaneporter:
- Added `RCloud.UI.session_pane.init();` to `mini.js`.
- `session_pane`'s `init` method assumed that there was a `#session_info` or `#output`. In mini/shiny, there was neither, but the code that later used this assumed there was a `#output` div. I have added a `.length` guard condition. If that is 0, then we have neither, so a fatal dialog is shown.
- `is_IE` function added which sniffs for IE. If IE is detected, a fatal dialog is shown immediately and execution returns. Unsure if this is the best/intended message for IE users. Perhaps the suggestion of using Firefox, Chrome (and others?) would be more useful.

### DIFF
--- a/htdocs/js/ui/session_pane.js
+++ b/htdocs/js/ui/session_pane.js
@@ -46,8 +46,8 @@ RCloud.UI.session_pane = {
     append_text: function(msg) {
         // FIXME: dropped here from session.js, could be integrated better
         if(!$('#session-info').length) {
-            console.log(['session log; ', msg].join(''));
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
I improved the formatting by moving the check to the beginning of the function, before the message has been wrapped in a jQuery error div.

gordonwoodhull:
I also changed the IE message to suggest using a different browser (and mention Edge). 

I still need to test this in IE, but it looks good.

PR #2000, opened 2016-05-31
Dialogs can be closed with 'esc'. Again.

### LINKED ISSUE #1999: can't close dialogs with ESC (yet again!)
I think this is the third time we've seen this, and each time a different reason. (Twice in #1863, once in #1928).

This time, it seems to be caused by 2a32bd2041b4b3df5b309bed3c6e97581b94cc5d "minor change to shortcut text" for cell deactivation.

Maybe a shortcut conflict? Do we need a global shortcut for cell blurring, considering that the cell should pick up ESC directly, considering it has the focus? The long, boring saga continues...

shaneporter:
This fixes #1999, which was caused by the blur cell/close dialog shortcuts using the same shortcut key.

In reality, they do not clash because they operate on different elements.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -192,7 +192,6 @@ RCloud.UI.init = function() {
         id: 'blur_cell',
         description: 'Blur Cell',
         keys: [
-            ['esc'],
             ['esc']
         ],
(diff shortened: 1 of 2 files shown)


gordonwoodhull:

PR #2002, opened 2016-06-01
Run all keyboard shortcut

### LINKED ISSUE #2001: Run All key
@shaneporter, sorry to add one last minute feature request, but it should be easy. Also, it's safe to do after the feature freeze if necessary.

Being the most obvious and most useful function, we should really have a key for Run All.

I really wish we could use a Function Key for this, but I know Windows [tends to gobble those up](http://ux.stackexchange.com/a/10698/83408), as does the debugger.

Ugh, let's punt and use ctrl/cmd-U (rUn, bleh) until we come up with something better. It's easy to change with feedback.

shaneporter:
Implemented in ACE editor and shortcut manager. This fixes #2001.

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -368,6 +368,16 @@ function create_cell_html_view(language, cell_model) {
             exec: function() {
                 ace_widget_.blur();
             }
+        }, {
+            name: 'executeAll',
+            bindKey: {
+                win: 'Ctrl-u',
+                mac: 'Command-u', 
+                sender: 'editor'
+            },
+            exec: function() {
+                RCloud.UI.run_button.run();
+            }
         }]);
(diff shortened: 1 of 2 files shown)


htdocs/js/notebook/cell_view.js
```diff
@@ -368,6 +368,16 @@ function create_cell_html_view(language, cell_model) {
...
+        }, {
+            name: 'executeAll',
+            bindKey: {
+                win: 'Ctrl-u',
+                mac: 'Command-u', 
+                sender: 'editor'
+            },
+            exec: function() {
+                RCloud.UI.run_button.run();
+            }
```
gordonwoodhull: Interestingly, this doesn't seem to be necessary - the global handler will still work if this is removed. 

I noticed this because I was concerned that the key would not work in the prompt or assets. But it does.

gordonwoodhull:

PR #2003, opened 2016-06-01
Feature/thumbnails

paulinshek:
Implement thumbnails feature

Replace previous calls to the github API to get the thumbnail by saving thumbnails separately in key-value database. This hopefully makes the discover page a bit faster.

Changes include two new R functions rcloud.get.thumb and rcloud.set.thumb. Only rcloud.get.thumb is exposed using RCAPS since rcloud.set.thumb hooks into rcloud.update.notebook

Resizing has not been implemented, so the missing.png has been resized manually.


paulinshek:
I forgot to say today: I've not implemented resizing of the images- I'm not sure where this would go? It would be more efficient in client side (and seems quite simple too: resize with canvas and then export with dataURI) but it means that the thumbnail code will be in multiple places?

Sorry for also committing the files created by grunt- I have added them to my gitignore now so shouldn't happen again.

Also, I resized the missing.png but then I thought that the original size might be deliberate... so am happy to change it back!

rcloud.support/R/zzz.R
```diff
@@ -7,4 +7,14 @@
...
+
+  printjs <<- rcloud.install.js.module("rcloud",
+                "(function() {
+                    return {
+                      console: function(message, k) {
+                        console.log(message)
+                      }
+                    }
+                  })()",
+                TRUE)
```
gordonwoodhull: We have `rcloud.session.log` for this (which prints to the session pane if available), so let's remove this.

rcloud.support/DESCRIPTION
```diff
@@ -5,7 +5,7 @@ Author: Simon Urbanek <urbanek@research.att.com>, Gordon Woodhull <gordon@resear
 Maintainer: Simon Urbanek <urbanek@research.att.com>
 Description: It's used by RCloud internally.
 Depends: base64enc (>= 0.1-3), rjson, parallel, R (>= 3.1.0)
-Imports: uuid, RCurl, unixtools, Rserve (>= 1.8-3), rediscc (>= 0.1-3), jsonlite, knitr, markdown, png, Cairo, httr, gist, mime, sendmailR, PKI
+Imports: uuid, RCurl, unixtools, Rserve (>= 1.8-3), rediscc (>= 0.1-3), jsonlite, knitr, markdown, png, Cairo, httr, gist, mime, sendmailR, PKI, digest
```
gordonwoodhull: looks like digest is not used anymore

rcloud.support/R/assets.R
```diff
@@ -63,12 +63,18 @@ rcloud.upload.asset <- function(name, content, notebook=rcloud.session.notebook(
...
     l <- list(list(content=content))
     names(l) <- name
     invisible(rcloud.update.notebook(notebook, list(files=l)))
 }
 
-rcloud.delete.asset <- function(name, notebook=rcloud.session.notebook())
-    rcloud.upload.asset(name, NULL, notebook)
+rcloud.delete.asset <- function(name, notebook=rcloud.session.notebook()) {
+    if (name=="thumb.png"){
+        rcloud.set.thumb(NULL, notebook$content$id)
```
gordonwoodhull: I think the parameters are backward here.

gordonwoodhull:
This looks mostly very good. The code is clean. I'm noting a few minor quibbles in the code in code comments. I would ordinarily just change myself, but since there are a couple of bugs, I'm going to ask you to change them instead.

It reduces the load time for 20 recent notebooks from 8 seconds to 3 seconds, so that's a pretty good improvement. (Of course I don't have any images cached, so I don't know if the speedup will be quite that good.)

Here are the bugs:
 - [ ] This does not allow deletion of thumb.png. The GitHub gist format indicates this by [passing `null` instead of the file object](https://developer.github.com/v3/gists/#edit-a-gist). It looks like the code now assumes there is an object there and replaces the `content` with the result of `rcloud.set.thumb`.
 - [ ] When I tried renaming the thumb manually in GitHub Gist, and then renaming it back inside RCloud to trigger the thumbnail code, I got the following error. I think this may be because the content is replaced with the result of `rcloud.set.thumb` which is the result of `rcs.set` which is probably something useless.

```
Error: rcloud.update.notebook: Validation Failed (422)
    at success (http://127.0.0.1:8080/js/rcloud_bundle.js:185:23)
From previous event:
    at rcloud_github_handler (http://127.0.0.1:8080/js/rcloud_bundle.js:188:24)
    at Object.rcloud.update_notebook (http://127.0.0.1:8080/js/rcloud_bundle.js:556:20)
    at update_notebook (http://127.0.0.1:8080/js/rcloud_bundle.js:3691:23)
    at Object.append_asset (http://127.0.0.1:8080/js/rcloud_bundle.js:3745:20)
    at http://127.0.0.1:8080/js/rcloud_bundle.js:1701:22
From previous event:
    at rename_file (http://127.0.0.1:8080/js/rcloud_bundle.js:1685:42)
    at execute_if_valid_else_ignore (http://127.0.0.1:8080/js/rcloud_bundle.js:1282:29)
    at HTMLSpanElement.keydown.editable (http://127.0.0.1:8080/js/rcloud_bundle.js:1294:32)
    at HTMLSpanElement.dispatch (http://127.0.0.1:8080/lib/js/jquery-2.1.1.js:4409:9)
    at HTMLSpanElement.elemData.handle (http://127.0.0.1:8080/lib/js/jquery-2.1.1.js:4095:28)
```

It would have been cleaner not to include the artifacts, but that's not a big deal. (It just makes the history a bit harder to read, which can lead to other mistakes.)

I am not sure if sizing the images matters, since masonry does this. I guess we could make them smaller to save space/latency. I don't know why missing.png was that size, maybe that's sized for retina?

rcloud.support/R/assets.R
gordonwoodhull: Asset deletion doesn't usually pass through here; I think this is an API for client code. Usually we'll send a batch of changes, including `filename: null`; I provided a link to the GitHub API in a comment.

htdocs/js/ui/discovery_page.js
```diff
@@ -23,59 +21,63 @@ RCloud.UI.discovery_page = (function() {
...
+                          return Promise.resolve({
+                            id: notebook[0],
+                            time: notebook[1],
+                            description: current.description,
+                            last_commit: new Date(current.last_commit).toDateString(),
+                            username: current.username,
+                            num_stars: editor.num_stars(current[0]),
+                            image_src: "data:image/png;base64," + thumb_src
+                          })
+                        });
```
gordonwoodhull: I don't think `Promise.resolve` is required here. A function returning inside a promise chain is always returning a resolved value.

gordonwoodhull:
Please push your fixes to the same branch and GitHub will pick them up in the PR. Thanks @paulinshek!

gordonwoodhull:
Hi @paulinshek, thanks for the updates. Is this ready for another review?

paulinshek:
Hi @gordonwoodhull Yep, all ready for another review now. I was able to reproduce all the issues that you mentioned- thanks so much! :)

rcloud.support/R/rcloud.support.R
```diff
@@ -344,6 +344,9 @@ rcloud.update.notebook <- function(id, content, is.current = TRUE) {
         content$files <- cfiles
     }
     content <- .gist.binary.process.outgoing(id, content)
+
+    # save thumbnails to key-value database
+    # rcloud.set.thumb(id = id, thumb_png = content$files$thumb.png.b64$content)
```
gordonwoodhull: accidental comment here, fixed in b77fecbf94

also it looks like you used a windows machine for this commit and it added some messy line endings, please take a look at https://help.github.com/articles/dealing-with-line-endings/#platform-windows

gordonwoodhull:
Rebased and merged. Thanks @paulinshek!

PR #2014, opened 2016-06-03
Feature/discover recent

paulinshek:
Change the "recent" ordering in the discover page to be dependent on last commit, rather than date created.

This uses a call to Sys.time rather than a call to github api for efficiency. However, this does mean that the date/time displayed might not exactly match that of github gist, but I think it will be close enough, maybe only a few seconds out.


gordonwoodhull:
Thanks @paulinshek! Unfortunately we still need the old "recent notebooks" feature since it's used to populate the menu on the notebook tree.

If you still have time to work on this today, please 
 - [ ] use a different RCS key like "recently modified"
 - [ ] introduce a new function to return the new data, and restore the old one to its old behavior

I realize the day is almost over there, so if there's no time, I can clean this up. I'm glad to see that the implementation is pretty straightforward.

gordonwoodhull:
Also, 
 - [ ] we have already called github and the modified date is in the result `res`

paulinshek:
Ahh, that makes sense! okay, will do!

paulinshek:
@gordonwoodhull I think it is working now... I'm having problems with my computer so haven't been able to fully test, but I have time to double check everything over the weekend as well!

gordonwoodhull:
That's okay, we're short on time so I'm just merging this and correcting stuff as I go. Thanks!

gordonwoodhull:
Rebased and merged. Thanks @paulinshek!

PR #2022, opened 2016-06-06
rcloud.get.notebook.forks

gordonwoodhull:
Hi @s-u, could you review this? I have never touched the gist codes. I think I'm following the pattern but I may have missed something.

Of course I did not implement this for git gists... I did have to add it to rgithub, thus the version bump and https://github.com/cscheid/rgithub/pull/65

We need this to implement the "most popular notebooks" feature.


gordonwoodhull:
Hmm, I forgot you're probably travelling, @s-u. And I want to go ahead and cut the release branch, so I'm going to merge this. Please review retrospectively.

Issue #1849, opened 2016-03-17
Unable to find the saved cell content with Find/Replace


sanketd11:
Steps to replicate:
- Add some content to a cell
- Invoke Find/Replace feature
- Try to search the cell content
- Save the cell contents
- Try again to search the saved cell content

Find/Replace feature will not highlight the content even though the content is present in the cell

shaneporter:
I'm seeing that after the find (or find/replace) feature is invoked, newly added content isn't 'found', regardless of whether that cell's content has been saved.

Only closing and reopening the find (or find/replace) feature 'recognises' the new text.

shaneporter:
Seems like there's an issue with the `content()` return value being 'stale'. I've managed to quite easily create something like this:

![image](https://cloud.githubusercontent.com/assets/2493614/14789873/8ea6ec3a-0b06-11e6-9f1d-f111f329505f.png)

gordonwoodhull:
Yeah, I don't think there's any attempt to keep the search results updated when the notebook changes. It's one of those loose ends I noticed right before a release and then completely forgot about.

Looks like the dirty bit handling in notebook_controller needs to provide an event that the find/replace logic can listen for. Then I guess the entire search needs to be recomputed. This might be slow, so a delay or throttling might be needed so that it's not doing a lot of work on every keystroke when the notebook is large.

shaneporter:
I don't know whether the search needs to watch for notebook changes. If I do a search right now on this page as I type, it matches what's currently there. If I type some more text, it doesn't update its search until I click the 'next' or 'previous' buttons.

gordonwoodhull:
This was not fixed by #2033

Issue #1839, opened 2016-03-14
Replaced word still getting highlighted in Find/Replace


Prateek032:
Steps to replicate
1. Create a cell which contains word repeated 2-3 times
2. Invoke Find/replace for that word
3. Now replace any occurrence of that word with the new word
4. Search for previous word again
Still the replaced word also gets highlighted for the search

gordonwoodhull:
Oo. Sounds bad. Thanks @Prateek032! 

This feature is a lot more complicated than it looks.

PR #2033, opened 2016-06-06
Find/replace matches/highlights issues fix

shaneporter:
An alternative approach to updating the search results as the notebook changes was suggested by @gordonwoodhull. Rather than incur an overhead in doing this, only show match highlighting when focus is on the find/replace fields.

As soon as a user clicks on a cell to edit, remove those highlights.

When the user clicks back into a find/replace field, reset the search using the current find/replace text.

This PR implements that functionality, and hopefully closes three bugs: #1839, #1849 and #1960.


gordonwoodhull:
This works great, however it only works when the find dialog loses focus to the cells in particular. 

So for instance you could start typing in the command prompt and generate more cells, or start editing the assets and generate more cells by executing code from there. It also seems more natural to me that the search results would disappear when editing a comment, searching or using help (focus leaving the notebook area).

Do the `focusin` and `focusout` events work on their own without the cell hook?

shaneporter:
I think you're right. There's no need for the specific cell hook. Can use focusout for the dialog. Will confirm in about 12 hours' time. 🙂

shaneporter:
Always good to have fewer lines of (clear) code to do the same things. Please review again @gordonwoodhull. Thanks.

gordonwoodhull:
So this does fix the old issues because it's impossible to save or modify the notebook, or open another notebook, without losing focus. However it creates one new issue #2052, and there is an old but I think unreported bug where it never found unsaved content #2053.

gordonwoodhull:

shaneporter:
As I was fixing #2052, I noticed that the 'focus out' approach doesn't catch everything. If you replace the text, the notebook's contents is modified without focus leaving the find/replace form.

PR #2010, opened 2016-06-02
Selected word should populate the find replace 'find' field; tab order on Find/Replace 'dialog'

shaneporter:
An implementation for #1846.

Fixes for #1984 and #1986, both of which are related.

I also include a commit ba563b8, which includes the find/replace close dialog's close button. Seems sensible to me, but easy to leave out if it's not wanted.

PR #2037, opened 2016-06-06
Selected word should populate the Find/Replace field (take 2)

shaneporter:
Previous PR (PR #2010) wasn't picking up new commits (cf18ab2 and 76a7bc3), guess this is because it was closed.

I created a reusable function in `ui_utils`, `copy_document_selection` that is now shared between document copy and find/replace 'find' population.


gordonwoodhull:
Great, thanks @shaneporter, works great - and sorry for the branch confusion.

PR #2041, opened 2016-06-06
'Run all' and 'Revert notebook' now available (once again) in view mode

### LINKED ISSUE #2021: almost all keyboard commands marked as write-only
The keyboard shortcuts dialog is pretty much empty for someone else's notebook:
image

I guess it makes sense that the code editor commands would be missing, but we need to go through and make sure that all the commands that should be available in readonly mode, are.

In particular, I miss cmd-U Run!

shaneporter:
Format of shortcut `keys` was incorrect, so it wasn't being picked up correctly. Changed this, and they now appear.

Having looked through the list, I don't *think* that there should be any more, since they relate to the ACE editor and associated shortcuts based on an editable notebook.

This fixes #2021.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -174,10 +174,14 @@ RCloud.UI.init = function() {
         category: 'Notebook Management',
         id: 'history_revert',
         description: 'Reverts a notebook',
-        keys: [
-            ['ctrl', 'e'],
-            ['command', 'e']
-        ],
+        keys: {
+            mac: [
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Ah, this may have been a bad merge - I was running into a lot of conflict on these kinds of lines.

gordonwoodhull:
I agree that the other commands are not relevant. It just surprised me how few there were, but I wasn't really thinking about it.

gordonwoodhull:

Issue #2043, opened 2016-06-06
Shortcut code is run multiple times


shaneporter:
I ran into this whilst fixing #2042; the first time a shortcut is used, its code is invoked once. The second time it's used, it's run twice. The third time it's used it's run three times. And so on.

This was causing an issue with #2042, and no doubt is causing some other issues too.

gordonwoodhull:
Whoah! That's crazy! 

Again, I think this was probably due to a bad merge because this code was worked over by multiple PRs (esp that longstanding ace keys PR and all the other shortcut stuff going on). Sorry about that, glad you found it quickly!

Issue #2042, opened 2016-06-06
populate find/replace from command prompt as well


gordonwoodhull:
as long as we still have a command prompt (#1950) invoking find from it should work the same as invoking find from a cell.

PR #2044, opened 2016-06-06
Prompt selection used for find/replace prepopulation and shortcut multiple invocation fix

shaneporter:
Whilst fixing issue #2042, I noticed something strange. Although I had modified the code, it wasn't working as I'd expected.

I found that shortcuts were being run _n_ times, where _n_ is the number of times that they have been invoked. :fearful:. I created a new issue #2043, and this PR fixes both the initial issue and #2043.

My logic for 'selection' is in this priority order:

1. Use cell selection from cell with focus. 0 or 1 of these.
2. Use prompt.

If neither have a selection, do nothing.


gordonwoodhull:

PR #2045, opened 2016-06-06
Select all in find/replace field not working

### LINKED ISSUE #2042: populate find/replace from command prompt as well

shaneporter:
I branched from my local bug/2042 branch, because that fixes #2042 and #2043. Without that, this was awkward to test.

The last commit 2f38893 fixes this specific issue. 

Default 'mousetrap' key handling prevents shortcut invocation from input fields. This can be overridden by adding a `mousetrap` class to the element. This was done for find/replace text inputs so that find and replace could be toggled whilst within the field.

This meant that the default browser implementation of select all no longer worked. 

I guess the special case for these fields has now propagated to the select all shortcut in `init.js`. Not 100% ideal, but it's a special case, and hopefully will be the only example of such.

### DIFF
--- a/htdocs/js/ui/command_prompt.js
+++ b/htdocs/js/ui/command_prompt.js
@@ -4,7 +4,9 @@ RCloud.UI.command_prompt = (function() {
         history_ = null,
         entry_ = null,
         language_ = null,
-        command_bar_ = null;
+        command_bar_ = null,
+        ace_widget_,
+        ace_session_;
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
Looks like I cherry-picked the commit on Monday but forgot to close this.

Issue #2032, opened 2016-06-06
Newly created notebook is not showing up in discover page


Prateek032:
The newly created notebook(empty notebook) is not displayed in Discover page. The notebook gets displayed only when the notebook contents are saved at least once

gordonwoodhull:
I argued in #2028 that showing new, unchanged notebooks is boring and useless. However, I guess it's surprising not to see them there, so we should show them anyway.

gordonwoodhull:
Fixed in #2047

amithrb:
This issue still persists on release-1.6/ 112bd78

gordonwoodhull:
I'm not seeing this currently.

amithrb:
@gordonwoodhull , Please try the following scenario:
- Refresh the URL of main page.
- Create new notebook.
- Go to discover page.

Unable to see the newly created notebook.
I can reproduce this issue on release-1.6/ 0cd33da.

gordonwoodhull:
Nope, I see the new notebook on the discover page.

Just to be sure, please verify the source of this function to make sure your server is up-to-date (sometimes it can be hard to completely kill and restart it):

![image](https://cloud.githubusercontent.com/assets/1366709/16330445/2dc27206-39b7-11e6-8243-b8f173a898dd.png)

gordonwoodhull:
`set.recently.modified.notebook` should do it.

gordonwoodhull:
Or if that checks out okay, please see if you can replicate the problem on rcloud-5, I just updated it.

amithrb:
I was able to reproduce it in rcloud-5.

gordonwoodhull:
I'm definitely not seeing it there, either. Is it possible that there is some odd time zone sorting problem and the notebook is actually displayed further down in the recent notebooks?

amithrb:
I have checked going further down in recent notebooks but, still not able to see it.
I will try to reproduce during WebEx call.

gordonwoodhull:
Okay, I am seeing this now, on a blank instance and on rcloud-5. 

There is no sign of some recently created or modified notebooks, even in the raw data returned to the page.

Being able to repro on a blank instance is very helpful.

gordonwoodhull:
I am no longer able to repro. On instances where there are a more than 100 notebooks, this could have been caused by #2136 which showed some arbitrary 100 notebooks in order of date, rather than the most recent.

#2154 describes an issue where the very first notebook created in a brand new DB does not appear.

[...]

Issue #2030, opened 2016-06-06
discover page should show number of forks


gordonwoodhull:
We haven't had this number available before, and we're not ready to show the actual list of forks, but since popularity is based on forks + stars, we should start by showing it on the discover page.

gordonwoodhull:
As noted in #2047, this is implemented (and the implementation looks correct at a quick read) but not quite working yet. Even when the fork count is updated by forking a notebook (and `forkcount` is populated in RCS), the count does not appear.

Issue #2028, opened 2016-06-06
discover page recently modified should include forks


gordonwoodhull:
it's not an update but `rcloud.fork.notebook` produces a new notebook.

gordonwoodhull:
no need to both with `rcloud.create.notebook`, though, since it will be completely boring if they haven't edited it.

gordonwoodhull:
Fixed in #2047

Issue #2011, opened 2016-06-02
renaming asset to thumb.png doesn't index the thumbnail


gordonwoodhull:
Although adding an asset named thumb.png does `rcloud.set.thumb` the thumbnail in the database, renaming an existing asset to thumb.png does not. 

This is strange because we implement rename as fetch-old, create-new, delete-old, so it should work the same.

I think a common workflow for adding thumbnails will be: take a snapshot, drag it to the assets, rename it to thumb.png, and that currently doesn't work.

gordonwoodhull:
(A little background: [GitHub Gist does have a rename mechanism](https://developer.github.com/v3/gists/#edit-a-gist), but [we aren't using it](https://github.com/att/rcloud/blob/develop/htdocs/js/notebook/asset_view.js#L18-L43). So I don't think this should be a special case, although I'm obviously missing something here.)

paulinshek:
thanks, will have a look!

paulinshek:
Hmm, it looks like renaming is not working properly, for me.

If I upload oldfile.png and then try to rename it to thumb.png, I get the following behaviour:
- Two new assets appear, both called thumbnail.png
- The content on one points to oldfile.png
- and the other points to thumb.png (but the gist is not modified, so I get a GET error)

I think there must be something missing from notebook.controller.append asset, or something else needs to also be called. I will keep digging.

gordonwoodhull:
Sounds like you're getting an error somewhere. The feature is well tested and should work. 

Temporarily displaying two assets is normal because of the silly way the feature is implemented, as described above. However, the old one should get removed after a moment.

gordonwoodhull:
Implemented in #2047. 

It's a really easy workflow on the Mac because you just hit cmd-shift-4, select the area on the screen, and then find the png on the desktop, drag it in and rename it. So it's great to see this working.

(Of course, typing all that out, it doesn't sound simple and we should really have pasting from clipboard to a new thumb.png asset, but I'm not sure how to do that.)

PR #2047, opened 2016-06-07
Enhancements and bug fixes to the discovery page

### LINKED ISSUE #2023: discover page shows zero stars for all notebooks
this information is not being shared; it's stored in editor_tab.js, which takes care of the notebook tree and the inventory.

### LINKED ISSUE #2039: Thumbnail image disappears from 'Discover' page after any changes in 'Assets' div
Steps :
- Create a new notebook and save after adding contents to it
- Upload thumb.png image to assets div
- Load Discover page (the thumbnail image will be displayed on Discover page for that notebook)
- Now load the same notebook again and make any change in assets div e.g edit Scratch.R, add new asset, delete asset, etc.
- Open Discover page (thumbnail image will disappear)

The thumbnail image will not be displayed even if we undo the changes made in 'Assets' div.

paulinshek:
Added an extra check that thumb.png.b64 is seen before testing for nullity, this should fix #2039, and #2011 

Discover page now shows the number of forks, to address #2030
- However, should clicking on the fork icon create a new fork of that notebook? I don't think clicking on the star adds to favourites?

Also, I have fixed #2023 and the star counts are now showing.

Finally, I also added calls to rcloud.config.set.recently.modified.notebook when a  notebook is forked or created, so now all newly created or forked notebooks show up on discover page recently modified,  to address #2032 and #2028


paulinshek:
Actually, the fork count is still always zero- will keep working on this!

rcloud.support/R/ocaps.R
```diff
@@ -137,7 +138,7 @@ unauthenticated.ocaps <- function(mode, compute)
       signal_to_compute = make.oc(.signal.to.compute),
       help = compute$help,
       get_thumb = compute$get_thumb,
-
+      get_fork_count = compute$get_fork_count,
```
gordonwoodhull: We should revisit if these need to be run from compute process or if they can be moved to the control process. If it works for now, that's fine, but I think it's generally better to put them on control if possible. (They will wait for cells to execute, and I think we eventually want to move all RCS access to control for security reasons.)

gordonwoodhull:
As for the fork count, this looks correct, pulling the count from RCS. The issue is probably that we need to populate `forkcount` with the value for all notebooks, since it's a new key and the source of the info is GitHub. 

We can create a notebook to do the population; the only tricky thing is running against [GitHub rate limits](https://developer.github.com/v3/#rate-limiting). Reading info about every existing notebook is a big deal.

gordonwoodhull:
In the meantime, you can test by refreshing the fork count on particular notebooks by.. forking them. They have to belong to someone else for it to be a "real fork", though, i.e. you may want a fake account.

rcloud.support/R/rcloud.support.R
```diff
@@ -407,14 +410,17 @@ rcloud.fork.notebook <- function(id, source = NULL) {
                                      list(owner=owner,
                                           description=src.nb$content$description,
                                           id=src.nb$content$id))
-    } else ## src=dst, regular fork
+    } else {## src=dst, regular fork
         new.nb <- fork.gist(id, ctx = src.ctx)
-
+        rcloud.config.set.recently.modified.notebook(id, new.nb$content$updated_at)
```
gordonwoodhull: This is creating an entry for the *source* notebook with the updated date of the *fork*. I'll open a new issue for this.

gordonwoodhull:
Great work, thanks @paulinshek!

For future commits, please use the phrases described here
https://help.github.com/articles/closing-issues-via-commit-messages/

in your commit messages, so that the issue will be automatically closed when the code is merged.

Rebased and merged.

paulinshek:
Thanks @gordonwoodhull, will do!

PR #2051, opened 2016-06-08
Fix #2024 Discover page thumbnails piling up

### LINKED ISSUE #2024: discover page thumbnails are piling up
I don't think this happened before #2014, although I don't see what changes there would have caused it. Seems like it is assuming all images are the same height, and a short height.

image

paulinshek:
Missing images loaded earlier so that they can be caught by imagesloaded, and before masonry initialises


gordonwoodhull:
Thanks @paulinshek! Rebased and merged

Issue #2034, opened 2016-06-06
New notebook is created with duplicate name


sanketd11:
Steps to replicate:
- Create a new notebook
- Add some contents to notebook and save or execute the notebook
- Open Discover page and click on the newly created notebook
- The respective notebook loads in RCloud 
- Now create a new Notebook

The new notebook will be created with duplicate name

gordonwoodhull:
Thanks @sanketd11! To be more precise, this will happen on the initial load of the page. 

@shaneporter, this is probably another case where the ids are wrong or we are not keeping all the notebook metadata that we used to. Try debugging `find_next_copy_name` and see why it is not finding that names are already taken.

gordonwoodhull:
Fixed by #2056

Prateek032:
This issue still reproducible, when notebooks are under any folder and try to create a  new notebook


![image](https://cloud.githubusercontent.com/assets/10941328/16292338/19bdf5de-392e-11e6-917a-af426ea6f373.png)

gordonwoodhull:
The root cause is almost certainly different.

I will bet that forking your own notebook does not produce a unique numbered name, if the source notebook is in a folder.

shaneporter:
I need a bit of clarification here please:

- when I create a new notebook using the + at the top right of the Notebooks panel, it creates a notebook in the root level of 'My notebooks'. Selecting a folder prior to clicking the + button makes no difference to this.
- when I fork a notebook (using the 'fork' button on the main navbar), the newly created notebook's name is unique, regardless of whether that notebook is in a folder or not.

Prateek032:
@shaneporter ,can you please try with the following steps:
-  add Notebook prefix (settings div) like  "folder_name/Notebook_name"
- Create  2-3 notebooks
-  Then refresh the tab and again create a new notebook
It starts again with the duplicate names

shaneporter:
Thanks @Prateek032! Yes, I see the duplication after the refresh.

gordonwoodhull:
This regressed again due to #2158.

PR #2056, opened 2016-06-09
New notebook names are unique (bug fix)

shaneporter:
This fixes issue #2034.

The find_next_copy_name function uses the alls/username part of the tree to test against code-generated notebook names for uniqueness.

The work for #1901 meant that this part of the tree might not be loaded, so checking against existing nodes was not particularly useful.

If this part of the tree has not been loaded, lazy_load will be true, and it is this that is tested to ensure that the children are loaded prior to the bulk of the find_next_copy_name function is run.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -468,44 +468,6 @@ var editor = function () {
         });
     }
 
-
-/*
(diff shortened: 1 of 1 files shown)


gordonwoodhull:

PR #2057, opened 2016-06-09
Focus out of find/replace form removes search matches

### LINKED ISSUE #2052: replace doesn't work unless you replace the first one
create a cell or notebook with a repeated word
 1. invoke replace for that word
 2. hit next till the second or later occurrence is has the current (orange) highlight
 3. hit replace (not replace all)

find appears to "reset": nothing is replaced and the current highlight goes back to the first occurrence.

the feature works if you replace the first occurrence.

shaneporter:
Apologies - the earlier commit, 707207d, did not fix #2052. The second commit, 7a71d05, should do.

I'm using `document.activeElement` to determine which element has the focus.

### DIFF
--- a/htdocs/js/ui/find_replace.js
+++ b/htdocs/js/ui/find_replace.js
@@ -41,12 +41,13 @@ RCloud.UI.find_replace = (function() {
                 generate_matches();
             });
 
-            find_form_.on('focusout', function() {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Nice. Thanks @shaneporter! 

(rebased to squash the earlier commit out of existence)

PR #2058, opened 2016-06-09
Find/replace matches content added after save 

shaneporter:
Small change to save the notebook when the `find_form_` gets focus.


gordonwoodhull:
Thanks @shaneporter!

PR #2059, opened 2016-06-09
Undo/redo available on edit, but not view page

### LINKED ISSUE #1899: Undo feature is not working for 'Read only' notebooks

shaneporter:
This fixes #1899 with the introduction of a new `on_page` property in shortcuts, defaulting to an array of `['view', 'edit']`. This property value is checked in the shortcut manager's `is_active` function against `shell.is_view_mode()`. 

Undo/redo shortcuts have their `on_page` property value set to `['edit']`. This means that they are only active on the edit page, and should only show in the shortcuts dialog on the edit page.

I cleaned up the 'revert' shortcut, which had a check for the page... this has now been moved into that shortcut's `on_page` property`.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -178,8 +178,9 @@ RCloud.UI.init = function() {
                 ['ctrl', 'e']
             ]
         },
+        on_page: ['edit'],
         action: function() { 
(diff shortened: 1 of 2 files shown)


gordonwoodhull:

PR #2060, opened 2016-06-10
Fix #2054

### LINKED ISSUE #2054: the fork source is being added as recently modified
It's a simple fix but passing this back to @paulinshek because i'm back in passive maintainer mode. 

The entry should use the new notebook's id instead of the fork source's id.

paulinshek:
Use the new notebook's id instead of the fork source's id when updated recently modified


gordonwoodhull:
rebased and merged. thanks @paulinshek!

Issue #2015, opened 2016-06-03
Abbreviate key commands 


gordonwoodhull:
It's a minor style point but I think the usual way to display key combinations is ctrl-cmd-alt-opt-shift-key. 

It will save a tiny bit of space too.

![image](https://cloud.githubusercontent.com/assets/1366709/15785362/db1ff3e4-2984-11e6-8ffd-208b4cd8c2f4.png)
looks too verbose to me.

shaneporter:
I was inspired by Github's help:

![image](https://cloud.githubusercontent.com/assets/2493614/15785635/37604372-29b0-11e6-87d6-0d521c877d75.png)

But I have no issue changing it.

shaneporter:
I've started looking at this, and I've abbreviated the mac keys. However, I don't believe using `-` over the current space will save any space.

I also think combining the keys in a single group conveys 'together' more effectively than using a `-`.

If you disagree, let me know, but I thought I'd let you know my opinion on it :smile:

gordonwoodhull:
Yes github is good precedent for sure. Thanks!

gordonwoodhull:
fixed by #2063

Issue #2006, opened 2016-06-02
second column of shortcut dialog does not display on smaller screens


gordonwoodhull:
we now have enough keys so that it's filling up and some of the keys are not visible unless you scroll.

but it looks like there's room for a second column of shortcuts.

this will really help when we merge #1903 ace shortcuts.

i'm not sure how hard this is (or if there's really enough room) so not putting it on a milestone yet

![image](https://cloud.githubusercontent.com/assets/1366709/15733358/3ee39efe-2853-11e6-8773-d0bfa7b15e55.png)

gordonwoodhull:
So yeah, I see that a second column does open up if the window is wide enough. Just not enough space on my macbook.

I fear that not everyone will find all the content if we have scrolling within a dialog, especially on a Mac where the scrollbars are hidden by default. But I know pagination is a lot of work for very little benefit. Maybe we can just make the content narrower for 1.6 so that it fits on smaller screens. Looks like it's the big gap between keys and descriptions which is causing this for me.

shaneporter:
What's the resolution on your laptop @gordonwoodhull? Just so I can see exactly what's going on.

Aside from anything else, the shortcut dialog looks ugly on your system. No so bad for me on Ubuntu, but clearly there are major improvements to be made. I think tabs by category would be a good way to go.

gordonwoodhull:
I've got a 2015 macbook so I think the emulated resolution is usually 1152x720. I don't think we'll have to go any smaller than that. 

Tabs rather than paging sounds like a good solution. Although I think we should be able to squeeze monow into the layout than we're getting now.

shaneporter:
Thanks. Does _monow_ mean more?

gordonwoodhull:
I have no idea what monow means 😄 but I meant more. Heheh.

shaneporter:
Thought it was an Americanism, but Google offered nothing. Thanks for the clarification. :smile:

shaneporter:
As I mentioned in #2015, I was inspired by Github's help dialog.

Because RCloud's main window doesn't scroll, I had to adopt a scrolling dialog body approach. Even with tabs, a category's content may not always fit within a fixed size body.

On non-macs, the scrollbar is fairly obvious, but perhaps not so much on OS X. Perhaps if I show a 'Scroll for more' message in the bottom right of the modal for a short period (or disappear when the content is scrolled)? Not ideal perhaps, but it means that:

- user will be aware that there is more content;
[...]

PR #2063, opened 2016-06-10
Shortcut dialog enhancements

shaneporter:
This addresses issues #2006 and #2015.

For shortcut dialog markup generation, I am now using an underscore template, which reduces the amount of boilerplate code in the dialog JS. Templates have been included in view/edit. Mac specific keys such as 'command' and 'option' have been shortened (	e6c20c1).

I have modified (and reorganised the SASS) so that there is room for 2 columns on smaller laptop screens. The first column of shortcut tables are now consistently sized, so the dialog content should look a lot nicer.

I had suggested showing a _there's more stuff, so you can scroll down a bit_ (bit pithy, but forgive me) indicator, but have, with commit 30acfb7, reinstated the vertical scrollbar so that it is always shown (props to http://simurai.com/blog/2011/07/26/webkit-scrollbar).


gordonwoodhull:
This looks great! 

I think the main thing that bothered me about the long modifier key names was that the key combinations were wrapping internally, making them hard to read. Each combo shows in a single line now.

Lots more information shown, and the scrollbar is clear enough. If we still get questions about this, we can add a "scroll me" overlay in the lower left but I don't think we'll need it.

gordonwoodhull:
rebased and merged. thanks @shaneporter!

PR #2064, opened 2016-06-10
Lose focus when cell header is clicked

shaneporter:
Lose focus of any element with focus, permitting delete shortcut to work as expected.


gordonwoodhull:
Oh that's nice and simple. I still want to remove the extra MVC events from #1991, since they are just noise now. I can do that. Thanks!

shaneporter:
Don't know why I didn't use it before!

gordonwoodhull:
Rebased and merged. Above commit removes the mvc blur stuff from #1991. ESC still works because it didn't need it.

Thanks @shaneporter!

PR #2065, opened 2016-06-10
Fixes #2024

### LINKED ISSUE #2024: discover page thumbnails are piling up

paulinshek:
Also move thumbnails code to discovery.R


htdocs/js/ui/discovery_page.js
```diff
@@ -35,7 +35,7 @@ RCloud.UI.discovery_page = (function() {
                                     last_commit: new Date(current.last_commit).toDateString(),
                                     username: current.username,
                                     num_stars: editor.num_stars(current[0]),
-                                    image_src: "data:image/png;base64," + thumb_src,
+                                    image_src: thumb_src,
```
gordonwoodhull: hmm, kind of looks like you didn't test this, since the base64 needs to be added back somewhere. i'll fix this.

gordonwoodhull: oh, i see, you added it to `rcloud.set.thumb` - that's reasonable but i'm going to keep that as raw data and assemble the url in the above line instead.

gordonwoodhull:
meh, please ignore my deleted comments. apparently we are using data urls rather than raw data for images when sending plots across the wire, so we should do the same here. good call.

gordonwoodhull:
rebased and merged. thanks @paulinshek!

PR #2067, opened 2016-06-13
New discover tab

### LINKED ISSUE #2061: Unable to access discover page as anonymous user
Following error message is displayed when discover page is accessed as anonymous user

image

As per our discussion, published notebooks should be displayed on discover page for anonymous user.

paulinshek:
Only loads what's needed for the discovery page.

I'm hoping that it also fixes #2061 and #2030, just double checking now.


htdocs/discover_tab.js
```diff
@@ -0,0 +1,58 @@
...
+        fork_count_ = {};
+
+    function clean_r(obj) { 
+        delete obj.r_attributes; 
+        delete obj.r_type; 
+        return obj; 
+    };
+
+    function load_everything(){
+        var allnotebooks = rcloud.config.get_all_notebook_info();
```
gordonwoodhull: This should be just the list of notebooks that are actually shown on the discovery page; loading the metadata for all known notebooks is prohibitively slow.

htdocs/discover_tab.js
```diff
@@ -0,0 +1,58 @@
...
+            notebook_info_ = clean_r(notebooks.infos);
+        }),
+        allnotebooks.then(function(notebooks) {
+            // populate star counts
+            num_stars_ = clean_r(notebooks.stars);
+        }),
+        allnotebooks.then(function(notebooks) {
+            // populate fork counts
+            fork_count_ = clean_r(notebooks.forks);
+        })];
```
gordonwoodhull: Doing `allnotebooks.then` on every branch probably shouldn't work. (If it does work, it's by accident.) Use `allnotebooks.then(function(...) { ... return Promise.all([...]); })` instead.

gordonwoodhull:
Putting this info into a class is a good idea; however, editor_tab has a number of anti-patterns which shouldn't be copied:

 - [ ] putting it in a separate file in `htdocs/` is an artifact going back to who-knows-what. this code could go better in the rcloud_bundle.
 - [ ] `_tab` is some horrible artifact from that won't die. I guess this is really the `discover_model`
 - [ ] the comment refering to model-model-view-controller doesn't really apply here (it's already pretty tenuous for the `editor_tab`)

This is all stylistic / code organization stuff having to do with general disorganization and cruft in the RCloud codebase; however, we shouldn't propagate these mistakes further.

gordonwoodhull:
All in all, this just needs a pass of cleanup and the efficiency gain of loading the precise notebooks we care about. Thanks @paulinshek, passing this to @shaneporter to finish while you're away.

htdocs/discover_tab.js
gordonwoodhull: oh, actually on further inspection, it can be just one callback, since none of the `then` functions are asynchronous themselves. but this code will probably change and look more like (but not exactly like) editor_tab's `get_notebooks_by_user`

htdocs/discover_tab.js
```diff
@@ -0,0 +1,58 @@
...
+            // populate notebook infos
+            notebook_info_ = clean_r(notebooks.infos);
+        }),
+        allnotebooks.then(function(notebooks) {
+            // populate star counts
+            num_stars_ = clean_r(notebooks.stars);
+        }),
+        allnotebooks.then(function(notebooks) {
+            // populate fork counts
+            fork_count_ = clean_r(notebooks.forks);
```
gordonwoodhull: `forks` doesn't exist in all_notebook_info, not that we should use it anyway. we'll probably need `rcloud.get_multiple_notebook_fork_counts` -- but that requires some server-side stuff. So if you'd rather not do that, @shaneporter, you could do something like 
```js
Object.keys(notebooks).map(function(id) { return rcloud.get_fork_count(id); })
```
for now. Slightly inefficient but shouldn't matter for 20 notebooks.

paulinshek:
I had added a fork count to get_all_notebook_info, I must have forgotten commit it. But it doesn't matter anyway, if we're not going to use get_all_notebook_info

htdocs/discover_tab.js
shaneporter: Hmm... perhaps I'm doing something wrong, but I'm getting null returned for get_fork_count.

gordonwoodhull: That's fine - we need to populate RCS, since this value normally comes from GitHub. You should plan for all the values coming up null initially, and you can populate some values by forking the notebooks of other people (causing an update).

We'll write a notebook that iterates through all current notebooks and populates RCS - but this is intense and will run against GitHub rate limits, so we have to be careful how we do it.

gordonwoodhull: obviously my demo code above needs a `Promise.all(...)` around it.

shaneporter: yeah :smile:

gordonwoodhull:
Replaced by #2077. Good start, @paulinshek, really set us in the right direction.

Issue #1787, opened 2016-01-26
Execute markdown cells when Notebook is opened


mudsam:
It would be very useful to have the option to have execute markdown cells when a notebook is opened.
I'm using markdown cells to have inline documentation of projects and being able to show the documentation upfront when the notebook is loaded would be great.

s-u:
The problem with that is that markdown cells can have side-effects so you cannot run them out of order without breaking the notebook. However, Marty mentioned that the underlying idea here was to have some kind of README cell at the top.

So I can think of a couple options:

 * have a special README.md asset which would be show at the top of the notebook - sort of like README on GH. It would re-render on commit.

 * we already have a "text" cell type, so we could add some kind of markdown "lite" (I shudder to think - we already have a proliferation of markdown-type cells) which would not allow side-effects. I'm not sure I like it - I guess a `html` cell would be somewhat less controversial as you know there will be no side-effect, but that may be less convenient for the author.

Comments welcome ...

mudsam:
I like the idea of a README, that's really what I'm trying to accomplish.

gordonwoodhull:
Okay, does this make sense: if there is an asset named README.md, it will get inserted as a special cell after the welcome message. I bet there is some way to ensure that knitr or rmarkdown does not interpret any R code, so we can basically use the existing interpreter with a special flag.

mudsam:
Yes, that makes sense, so you would use knitr with eval=FALSE to render it?

holyguster:
I like, but... I go back and forth about whether I like the asset idea, as that's really a divergence from the norm.

If sourced from an asset, it needs to be really clear that the contents of the special cell after the welcome message came from the README.md asset.  Maybe it's title could be "Asset README.md" or something like that?  

Also, that special cell shouldn't have any of the cell buttons (split, combine, edit, etc), and shouldn't be grabable/moveable by clicking in the left gutter.  Probably not hard, as all that is similar to when viewing a notebook owned by someone else.

How can it be made clear to the notebook author that R code won't be run, and why?  Maybe using .html instead of .md (as Simon suggested) would handle the first part.

On second thought, if the README.md is always executed as the first cell, are side effects still a concern?

I definitely like this idea, but still chewing on it...

gordonwoodhull:
Yeah, all that stuff is complicated. Probably worth it, but complicated. 

Plus: while you are editing the markdown in the asset, when it autosaves it will update the notebook area. It has to.

[...]

PR #2077, opened 2016-06-15
Discover fixes

shaneporter:
This should tackle most of the remaining 1.6 Discover issues, except #2030 (I am getting an array of nulls returned for the forks), so the fork count is hard-coded at 1. 

Generally speaking, there are improvements to performance (though I'm not able to verify just how much with a small local instance), UI and browser _bookmarkability_.


gordonwoodhull:
Wow, this is huge.

gordonwoodhull:
Oh wow, I love the new look and background!

shaneporter:
Apologies it's so big. I went on a bit of a rampage last night.

In future, it'd be nice to be able to add a description to a notebook, so that it can be picked up here.

gordonwoodhull:
Not a problem at all -  I appreciate a good rampage.

Re: the description, we're actually misusing the description tag as the name of the notebook. (Gists don't have names, github just uses the first filename as the "name".) 

Maybe if we adopt the README cell #1787 (remember that?) we can show an excerpt (title or first paragraph) on the discover page.

gordonwoodhull:
This is just about perfect. I'm going to merge and run with it from here.

gordonwoodhull:

PR #2094, opened 2016-06-16
Discover 'tiles' don't overlap

shaneporter:
So, this isn't the first time a 'fix' has been submitted for #2024.

I was able to reproduce the issue, but only intermittently, and infrequently enough that it was understandable that I hadn't seen it for a while. 

Try as I might, I haven't been able to reproduce it again since 

`missing.png` was causing the issue, and I suspect that a caching issue was causing the problem and that imagesloaded's callback was firing too soon. Seems as though [there are others that have faced this issue](https://github.com/desandro/imagesloaded/issues/103). I updated imagesloaded _and_ masonry, but this didn't help. 

I had toyed with the idea of adding a hidden `missing.png` to the discover page with a width and height of 1 pixel. But I didn't.

Instead, I converted the `missing.png` image to base64, added it to the discovery_page, and used that instead. It's not my preferred solution, but I _think_ it works.

(I also noticed that at widths < 768 pixels, the nav height increased, so added a media query to push the body down a bit.)


gordonwoodhull:
Rebased and merged. Thanks @shaneporter! I'll comment more on #2024 (which I'm not able to repro).

PR #2095, opened 2016-06-16
Cannot import notebook from file without description

shaneporter:
Although the first commit fixes the issue, it didn't make the 'import' button behave in a consistent way to the other 'validation' concerning the presence of a file.


gordonwoodhull:

PR #2096, opened 2016-06-16
Read-only notebooks assets stay as read-only

shaneporter:
Added a small amount of code that updates the ACE widget if the notebook is read-only.


gordonwoodhull:

PR #2097, opened 2016-06-16
If it's a help pane, it grows

shaneporter:
Seems to fix it - previous code assumed a left panel help layout. Now uses a bit of jQuery to ascertain where it is and calls the correct code.


gordonwoodhull:
Cool. We'll probably want to put this in a reusable function later. (I think the only other panel that does this is the session pane, which doesn't currently change sides.)

gordonwoodhull:
Except... why is this working for me even without the fix?

Not worth questioning. Rebase, merge, forget, move on.

gordonwoodhull:
Thanks @shaneporter!

Issue #2090, opened 2016-06-16
Some of the Keyboard short cuts are not working


Prateek032:
Following are some of the keyboard shortcuts are not working.

These we "stole" from Ace and we would probably give them back.
- [x] Ctrl U / Ctrl shift U -changing to upper/lower case
- [x] alt up / alt down - Move lines up/down

This is the browser's and [we intentionally disable the Ace behavior so as not to interfere](https://github.com/att/rcloud/blob/develop/htdocs/js/ui_utils.js#L159) - just need to remove the description:
- [x] Ctrl L - go to line

This is impossible because Ace no longer scrolls in cells. We either should scroll the notebook or we should use this for next / last cell:
- [x] Ctrl up/ Ctrl down - Scroll lines up/down
 
This should work:
- [x] cmd shift i - Invert selected cells

and when we open any other users notebook, 
- [x] ctrl e is listed

Revert should only be available if the notebook is yours and a historical version is open.

gordonwoodhull:
Wow keys are a tight resource. I've done some triage above. @shaneporter, please comment with any questions or suggestions.

gordonwoodhull:
Thanks @Prateek032!

shaneporter:
So, these fall into categories:

1. Not clashing and not working, so fix (_invert selected_, _revert_)
2. Remove (ctrl up/ctrl down) and goto line.
3. Change clashing shortcuts keys for _run all_ and _move lines up/down_

Suggestions:

run all: ctrl shift enter
go to previous/next cell -  We can't use _ctrl/cmd left/right_ for that because that's been taken by ACE.  

I'll start looking at the non-working shortcuts.

gordonwoodhull:
@shaneporter, I agree with your conclusions. I've checked off the stuff implemented in #2101 and I'm also checking off cmd/ctrl-E as wontfix, because the conditions where Revert is applicable are so unique that it would be weird for it only to show up in the shortcuts help in those circumstances.

It's kind of odd that Ace doesn't assign a key for scroll line down on Mac; let's just go ahead and remove those descriptions. Unfortunately we can't use ctrl-up and ctrl-down for last/next cell because those do window management stuff on Mac. I still don't know how to resolve that.

ctrl-shift-enter is brilliant for Run All!

gordonwoodhull:
Fixed by #2103

PR #2101, opened 2016-06-17
Shortcut fixes for 1.6

shaneporter:
#2090 has a few sub-issues, and although this PR doesn't resolve all of them, it does implement changes that fix all but the 'shortcuts we stole from ACE' items.

Submitting now so that some of those issues can be marked as done.


gordonwoodhull:

PR #2105, opened 2016-06-20
Can now add members/admins to group

shaneporter:
`editor.allTheUsers` _was_ used to initialise the users. Although that sounded about right, its value was undefined, and there was no code to get the users.

I have added a new scope level function to get the users. This wasn't enough, however, since I couldn't get things to work until I converted the names to selectize compatible format.

Becuase both selectize directives use the scope level `allUsers` member it works for both admins and members.

### DIFF
--- a/htdocs/js/ui/notebook_protection_controller.js
+++ b/htdocs/js/ui/notebook_protection_controller.js
@@ -52,6 +52,9 @@ define(['angular'], function(angular) {
             //general logic
             $scope.evalData()
             .then(function() {
+                return $scope.getUsers();
+            })
+            .then(function() {
                 return $scope.getGroups();
             })
             .then(function() {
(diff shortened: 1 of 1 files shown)


htdocs/js/ui/notebook_protection_controller.js
```diff
@@ -94,6 +97,24 @@ define(['angular'], function(angular) {
...
+                .then(function(data) {
+
+                    $scope.$evalAsync(function() {
+                        $scope.allUsers = _.map(data, function(user) { return { text : user, value : user }; });
+                    });
+
+                    //timeout to ensure evalAsync has finished
+                    $timeout(function() {
+                        logger.clear();
+                        resolve();
```
gordonwoodhull: Hmm, is async-call-then-timeout-resolve really the right pattern? Seems like we should just resolve once the asynchronous call is done.

gordonwoodhull: Just a question, I'm going to merge this.

gordonwoodhull:
Ah, okay, this makes sense now - thanks for the explanation. 

Yes, I suggested `allTheUsers` to the previous author of the discovery page as a temporary hack. You removed it in a1f775a01d4 as part of simplifying `load_everything`, which was a good call. It's surprising that it failed so quietly but I'm glad to remember how this used to work.

htdocs/js/ui/notebook_protection_controller.js
shaneporter: It's smells a bit doesn't it? Having said that, I took the easier option, and duplicated the issue from other areas of the code. Perhaps it's something that can be looked at in the future.

gordonwoodhull: #2114

gordonwoodhull:

Issue #1956, opened 2016-05-05
"aw shucks" is white in shiny.html


sanketd11:
![capture](https://cloud.githubusercontent.com/assets/9251570/15036647/dd783570-12af-11e6-8f72-2495408b56a6.PNG)

gordonwoodhull:
Hmm, adding rcloud.css to shiny.html breaks other things. There is a reason we didn't. 

Notice the fatal dialog alignment is off and it also makes one of the select elements huge.

We probably would need a minimal CSS that just styles this dialog.

![image](https://cloud.githubusercontent.com/assets/1366709/15074489/58e4927c-136d-11e6-9b0d-3effccb3ae71.png)

shaneporter:
What's the simplest shiny notebook I can use to replicate this issue? When I invoke the fatal dialog on shiny.html using `RCloud.UI.fatal_dialog`, I see this:

![image](https://cloud.githubusercontent.com/assets/2493614/15109683/8e14dd3a-15d4-11e6-821d-0af9f25f7401.png)

I'm not concerned about the `undefined`, but there is _no_ styling because no CSS resources are requested. Perhaps I'm doing something wrong.

sanketd11:
Try with following steps:
- Open any shiny notebook with Shiny.html  
- Kill the Rserve
You will get the dialog this way

shaneporter:
@sanketd11 thanks, I can simulate an issue by invoking the fatal dialog from the JavaScript console. I was wondering about a minimal shiny notebook. Do you have one you could attach?

gordonwoodhull:
Was about to write the same thing. 

Shiny will dynamically load the bootstrap css and a bunch of other dependencies, so that might explain the unformatted output you saw.

gordonwoodhull:
Can't think of a fatal error that will occur from within Shiny, we pretty much just let them take over.

shaneporter:
thanks @gordonwoodhull can you please attach an exported notebook that you used for the screenshot above?

gordonwoodhull:
Better, all the supported Shiny notebooks are on rcloud.social and you can use the same notebook ids on your own instance if public github is your notebook store.

http://rcloud.social/edit.html?notebook=68d114c159e5c364ae7e#

shaneporter:
[...]

PR #2106, opened 2016-06-20
Shiny 'Aw, shucks' looks nicer

shaneporter:
I was about to open a PR and remembered that the first commit would cause an issue as reported in #1956.

I created a 'special' RCloud shiny SASS file and scoped things within the dialog, so they shouldn't interfere with other Shiny elements.

### DIFF
--- a/htdocs/lib/js/require-shiny.js
+++ b/htdocs/lib/js/require-shiny.js
@@ -13,15 +13,17 @@ requirejs_config_obj = {
             deps: ['jquery'],
             exports: "jQuery.fn.dataTable"
         },
+        "jquery-ui": ["jquery"],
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
The dialog still shows up off to the right when Shiny has already loaded, but it looks good.

shaneporter:
Sorry it's off to the right in that scenario - is it easy to reproduce?

gordonwoodhull:
Yeah, just kill the server after the shiny page has loaded. 

Another glitch is that we should report a nicer error for the "not published" message, as we do in view.html.

Neither of these really matter, though. Just sort of documenting the glitches here.

shaneporter:
OK, thanks.

Issue #2082, opened 2016-06-16
find should current-highlight first found


gordonwoodhull:
find uses an orange highlight to indicate the current match.

as the search is executing, it should make the first match current.

another symptom is that currently searching in reverse will first current-highlight the first and then highlight the last on the second click. if the first had already been current this would work correctly.

gordonwoodhull:
fixed by #2107

gordonwoodhull:
failed to merge somehow

Issue #1720, opened 2015-10-21
Enhancement: Find and Replace feature  should display count for search results


Prateek032:
Currently it highlight the results but it doesn't display the count for the search results. As shown below

![find_replace](https://cloud.githubusercontent.com/assets/10941328/10630429/67eaa172-77a4-11e5-8067-a2abd4398f92.PNG)

shaneporter:
Chrome's search automatically highlights a matching string, which is only done in RCloud when the 'next' or 'previous' buttons are clicked.

The match index is also performed from the current selection (if any).

So, to do what Chrome does, we would need to:

1 - Perform the initial highlight when search text is modified.
2 - Calculate the match index based on the current selection (if any).

Seems as though they are desirable improvements that could be encapsulated by this enhancement.

gordonwoodhull:
I am seeing it highlight immediately, but it doesn't highlight the "current" one in orange immediately, or scroll to it. IT

shaneporter:
Suggested implementation opened with #2107.

gordonwoodhull:
implemented in #2107

Prateek032:
I couldn't find this enhancement on release-1.6/ 112bd78

gordonwoodhull:
Thanks @Prateek032, failed to merge this somehow, must have been juggling branches. 

Reopening to close with proper commit.

PR #2107, opened 2016-06-20
Find/replace match count (and other fixes)

shaneporter:
I have added code to display match count within find/replace (#1720) and auto-highlight first match (#2082).

If replace is invoked (using the shortcut) after the find is shown, focus will be set to the replace field.


gordonwoodhull:

gordonwoodhull:
Too many branches, let's try this again.

PR #2108, opened 2016-06-21
Discover Metric change indicator

shaneporter:
The existing loader has been modified so that it overlays the entire screen. This prevents the user clicking the metric links whilst things are loading.

Scroll to top code has been implemented too, so the user doesn't have to scroll to the top manually after the metric has changed.

Locally, it's quite fast, so to simulate a long-running process, I added a setTimeout. Removed it in 	4007a06.

### DIFF
--- a/htdocs/discover.html
+++ b/htdocs/discover.html
@@ -21,8 +21,10 @@
     </div>
 
     <div id="progress">
-      <img src="img/logo_rcloud.png" />
-      <div class="loader">Loading...</div>
+      <div>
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
The smooth scroll seemed to be causing flicker (the tiles drawn twice?) so I switched to a sudden scroll after loading the metric. I guess the loader is enough indication that something is changing.

gordonwoodhull:

shaneporter:
Fair enough.

PR #2109, opened 2016-06-21
Shortcuts disabled when error occurs on view

### LINKED ISSUE #1873: Error when <DEL> key is pressed in anonymous view.html

shaneporter:
This fixes #1873. Added a 'disable all' method to the shortcut manager.

This method is invoked when a serious error occurs on view/edit.


gordonwoodhull:
Rebased (squashed) and merged. Thanks @shaneporter!

PR #2110, opened 2016-06-21
Renaming parent folder of current notebook in notebook tree updates title

shaneporter:
Update notebook title when parent folder contenteditable is updated.


gordonwoodhull:
Thanks for tracking this down, went higher level and used `rename_current_notebook` from the current scope instead. (It's still your fix!)

shaneporter:
Cheers!

PR #2120, opened 2016-06-23
Command prompt blur on escape

shaneporter:
Took what was in cell_view and propagated it to command_prompt.

Modified the 'esc' description to reflect the command prompt blurring.


gordonwoodhull:

PR #2137, opened 2016-06-27
Top 100 not First 100 on Discover

shaneporter:
Changed ordering of code so that slicing occurs after sorting.


gordonwoodhull:
Awesome! Seeing some genuine popularity on rcloud-5 now - scores up to 46 (forks+stars) and all the top ones are business-related rather than just demo notebooks.

Thanks @shaneporter!

PR #2138, opened 2016-06-27
Load notebook by ID now works with more forms of URL

shaneporter:
Modified code so that this URL can now be imported:

`https://gist.github.com/d90a67db7559554941f032a48b8febf8`


gordonwoodhull:

PR #2141, opened 2016-06-27
View mode 'find' now behaving itself

shaneporter:
Improved the look of the find input area.

Also fixed the issue, which was related to an undefined `RCloud.UI.command_prompt.ace_widget()`.

(So I think this was unrelated to very recent changes.)


gordonwoodhull:
It was related to even more recent changes: I apparently broke it in 84aa604a10e, when clarifying the logic to find text to populate the find/replace dialog. Thanks!

gordonwoodhull:

PR #2145, opened 2016-06-28
Default panel size setting does agree with actual layout

shaneporter:
Perhaps this doesn't need the _.isUndefined(...) because after running 

````r
rcs.rm(usr.key(notebook='system', 'config', 'panel-layout-by-size'))
````

I get `null` as the resolved value.

### DIFF
--- a/htdocs/js/ui/panel_loader.js
+++ b/htdocs/js/ui/panel_loader.js
@@ -197,7 +197,13 @@ RCloud.UI.panel_loader = (function() {
             }
 
             // alternative layout?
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I see. Design-wise, it's a consequence of having the settings frame load the settings, while the panel loader which needs this setting is the one that loads the settings frame, so it's too early to take advantage of any defaulting. This wouldn't be necessary if there were a settings model distinct from the frame. 

Ah well.

gordonwoodhull:

PR #2151, opened 2016-06-29
Solr delete hidden notebooks

prateek05:
Closes #1078

### DIFF
--- a/rcloud.support/R/rcloud.support.R
+++ b/rcloud.support/R/rcloud.support.R
@@ -495,8 +495,14 @@ rcloud.is.notebook.visible <- function(id)
     visibility
   }
 
-rcloud.set.notebook.visibility <- function(id, value)
+rcloud.set.notebook.visibility <- function(id, value){
   rcloud.set.notebook.property(id, "visible", value != 0);
+  if(value){
+    response <- update.solr(rcloud.get.notebook(id),rcloud.notebook.star.count(id))
+  } else {
+    response <- .solr.delete.doc(id)
+   }
+}
 
 rcloud.port.notebooks <- function(url, books, prefix) {
   foreign.ctx <- create.github.context(url)
--- a/rcloud.support/R/notebook.protection.R
+++ b/rcloud.support/R/notebook.protection.R
@@ -42,6 +42,10 @@ rcloud.set.notebook.cryptgroup <- function(notebookid, groupid, modify=TRUE) { #
                          stop(e) })
         }
     }
+    ## Remove notebook from SOLR Caching notebook was successfully assigned a crypto group
+    if(is.notebook.encrypted(notebookid)) resp <- .solr.delete.doc(notebookid)
+    ## If group change is set back to Public update the solr index
+    if(!is.notebook.encrypted(notebookid)) resp <- update.solr(rcloud.get.notebook(notebookid),rcloud.notebook.star.count(notebookid))
     invisible(TRUE)
 }
 
--- a/rcloud.support/R/notebook.protection.R
+++ b/rcloud.support/R/notebook.protection.R
@@ -46,7 +46,7 @@ rcloud.set.notebook.cryptgroup <- function(notebookid, groupid, modify=TRUE) { #
 }
 
 is.notebook.encrypted <- function(id)
-  sapply(property.multiple.notebooks(id, "cryptgroup"), function(x) !is.null(x))
+  sapply(property.multiple.notebooks(id, "cryptgroup"), function(x) {!is.null(x) && (x==TRUE)})
 
 rcloud.get.cryptgroup.users <- function(groupid) { # : list(user -> is.admin)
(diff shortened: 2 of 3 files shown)


gordonwoodhull:
Fantastic, this is a big relief! Code looks good.

It happens this will also unindex deleted notebooks, because I started setting them hidden before deleting. (It's the easiest way to exclude them from the discovery page.)

Can I ask you also to extend it to also exclude encrypted notebooks from search? Right now they are indexed up until they become protected, and then they remain as stale results (and clicking them produces an error).

I think this is just a matter of pulling the add/remove logic into a new function, and calling it from `rcloud.set.notebook.cryptgroup` as well. (Protected notebooks aren't hidden because that wouldn't account for protection groups.)

prateek05:
yes i can look into that

gordonwoodhull:
Thanks! There is also `is.notebook.encrypted` to check whether an update should be indexed or ignored.

I thought about generalizing all this (pub/sub events on notebooks, sort of) when I implemented the same sorts of checks for the discovery page in discovery.R. But eventually I had to just get on with it.

I am very grateful that you're fixing these parts of search. It's been embarrassing to see results pop up that shouldn't.

prateek05:
@gordonwoodhull I rebased release/1.6 commits on the current solr patch and i see additional commits from release 1.6 appearing as part of this PR.
In such a situation do I need to revert back or is this PR safe since its raised against release/1.6

prateek05:
I ll revert back and apply only my commits. The difference is accounting for all of the changes and I am not sure now if some other commit "potentially" creeped in

prateek05:
@gordonwoodhull PR fixed.

rcloud.support/R/rcloud.support.R
```diff
@@ -495,8 +495,14 @@ rcloud.is.notebook.visible <- function(id)
     visibility
   }
 
-rcloud.set.notebook.visibility <- function(id, value)
+rcloud.set.notebook.visibility <- function(id, value){
   rcloud.set.notebook.property(id, "visible", value != 0);
+  if(value){
+    response <- update.solr(rcloud.get.notebook(id),rcloud.notebook.star.count(id))
```
gordonwoodhull: This is crashing on binary assets, but I think I can handle it somehow.

gordonwoodhull: Looks like what we are doing in other cases is indexing the base64-encoded gunk, haha. So I guess by that standard, we can just fetch with raw=TRUE. But that's pretty terrible...

prateek05: Should we check for binary assets in the update function and prune these assets out from checking in ? This would have never worked ever since the solr schema has a data type of string. 
Or we go with the base64encoded string ?

prateek05: b5c282922a3325b59e633c45c5f01a0a4ab6eb76 should fix the R type 24 to JSON error in binary assets

gordonwoodhull:
Hmm, I was thinking that we don't need to index those files at all, so we could just screen out anything with the .b64 extension.

gordonwoodhull:
BTW I was finally able to test this on rcloud-5. It works great!

prateek05:
nb$content$files$zip.jpg i dont see the extentions as .b64 the notebook object has the original extention of the file

gordonwoodhull:
The extension will be present in the raw version, which is what we send to GitHub.

prateek05:
how about we check the content of the file and if it returns "raw" dont check it in ? Would that work ?
mode(nb$content$files$zip.jpg$content)
"raw"

prateek05:
let me check with the extention in the raw version i should be able to do this now

gordonwoodhull:
I think the other places we call `update.solr` we are passing the raw version, so it wouldn't work there. We should consistently do one or the other.

gordonwoodhull:
E.g. in `rcloud.update.notebook` we're passing the result of `.gist.binary.process.outgoing`, i.e. the raw and/or encrypted version.

prateek05:
Right! Do we want to close this PR? I ll open up a new Issue with improvements for removing binary assets while checking the notebook in solr ? We could discuss that further in that work item. I am guessing we need update solr to handle both raw and the github format ?

gordonwoodhull:
Yeah, I've already merged this actually. (I have to close them manually when I rebase and forgot here.)

I think it should be pretty simple. Just filter by filename. The raw format is just the github format with each binary asset base64 encoded and renamed.

gordonwoodhull:
Then just make sure every place `update.solr` is called, it's the raw format.

prateek05:
Quick question though I just checked rcloud-5. 
Commits 75bb3a4 b5c2829 5b9a41a are missing.

gordonwoodhull:
Yeah, I rebased and removed the commits that were false leads. So it's just the relevant stuff.

prateek05:
Ah i see

PR #2156, opened 2016-07-01
Discover small thumbs and fork/star titles

shaneporter:
Changed the background of the tile so it looks better when the thumb's width is less than that of the tile. 

image

Made the entire tile clickable.

Also changed the fork/star titles because they (especially the star) implied that something could be done by clicking them (over and above opening that notebook.)


gordonwoodhull:
Looks great! Thanks @shaneporter, rebased and merged.

PR #2157, opened 2016-07-01
Don't always scroll to current notebook

shaneporter:
There are several situations where selecting a node should mean that it scrolls into view.

The lazy loading for children should not need to scroll because a different node has not been selected.

Updated the `select_node` function so that by default it scrolls by default. If passed a falsey value for the `scroll` parameter, it does not scroll. A `false` value is passed in when called from the `load_lazy_children` function.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -837,9 +837,14 @@ var editor = function () {
         ui_utils.scroll_into_view($tree_.parent(), 50, 100, $(node.element));
     }
 
-    function select_node(node) {
+    function select_node(node, scroll) {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks! 

I failed to mention that I started working on this in a branch. It turned into a major refactor (better following the MVC pattern), and I'm a little worried it might not resolve, so it's great to have this as a fallback! I'll know later today which route to take.

shaneporter:
Oh ok - sounds good.

gordonwoodhull:
Went with a comprehensive approach in #2158

PR #2163, opened 2016-07-02
htmlwidgets: remove periodic resize checker, closes #2162

### LINKED ISSUE #2162: htmlwidgets are growing
Some htmlwidgets, in particular networkD3 and dygraphs, grow about 1cm taller every 5 seconds.

gaborcsardi:
It might result an infinite resize loop for some widgets. Hopefully this solves #2162.

### DIFF
--- a/rcloud.support/R/htmlwidgets.R
+++ b/rcloud.support/R/htmlwidgets.R
@@ -78,6 +78,12 @@
 ##
 ## # Sizing
 ##
+## [ NOTE: Currently we turned off the periodic size check that ran every
+## five minutes. It generates an infinite loop for some widgets:
+## plotly, networkd3 and dygraphs. In theory this might mean that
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @gaborcsardi. 

That's too bad that the sizing might be off sometimes, but it sounds like the right call for now.

Rebased and merged.

gaborcsardi:
I talked to Joe, JJ and Ramnath about this last week, and they didn't have a better solution, either. :(

gordonwoodhull:
I guess if we could eliminate the iframe that would solve it, but who knows what else that will entail. 

I know that rmarkdown will generate html fragments via pandoc, but I don't know if htmlwidgets can be coaxed to do the same. (Also it means injecting unknown JS and CSS into the page.)

Anyway this is a great start.

PR #2164, opened 2016-07-06
"Fix" htmlwidgets sizing

gaborcsardi:
This adds back the periodic resize checks, but widgets are not getting bigger any more, except if the window or cell is resized to be bigger.

Hopefully solves the problems as much as possible. I have tested it with "good" and "bad" widgets, and also in mini.html.

### DIFF
--- a/rcloud.support/R/htmlwidgets.R
+++ b/rcloud.support/R/htmlwidgets.R
@@ -78,12 +78,6 @@
 ##
 ## # Sizing
 ##
-## [ NOTE: Currently we turned off the periodic size check that ran every
-## five seconds. It generates an infinite loop for some widgets:
-## plotly, networkd3 and dygraphs. In theory this might mean that
-## some widgets might not size properly, especially in mini.html.
-## A possible workaround is to resize the window (again) by hand. ]
-##
 ## Sizing of html widgets is tricky by itself. See the vignette in the
 ## R package, currently here:
 ## https://cran.r-project.org/web/packages/htmlwidgets/vignettes/develop_sizing.html
--- a/rcloud.support/R/htmlwidgets.R
+++ b/rcloud.support/R/htmlwidgets.R
@@ -202,14 +202,14 @@ rcloud.htmlwidgets.viewer <- function(url, height) {
   invisible()
 }
 
-as.character.htmlwidget <- function(x, ...) {
+as.character.htmlwidget <- function(x, ocaps = TRUE, ...) {
   tmp <- tempfile(fileext=".html")
   on.exit(unlink(tmp), add = TRUE)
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Wow, but the plots get a **lot** taller when you resize the window. I think it may be seeing multiple resize events?

before window resize
image

after
image

We may have to roll with this as it is, but please check if you're seeing multiple events, and debounce if so.

gaborcsardi:
OK, I'll give it another go. For me it wasn't that bad, actually. 

I guess I can just throttle the resize events, or only execute the last one, if no others are coming for (say) a second.

shaneporter:
Could [this](https://davidwalsh.name/function-debounce) be useful?

gordonwoodhull:
Yeah, I was thinking of the underscore throttle or debounce, thanks!

Actually, I guess this behavior existed before this PR (it's observable on rcloud.social without this merged), but I'll still wait on merging this.

shaneporter:
I seem to recall using [this](http://marcj.github.io/css-element-queries/) in the past, and found it to be an effective way of applying dynamic CSS classes based on element width. 

Could that ResizeSensor demo on that page be useful?

gaborcsardi:
OK, we worked it out with Shane and it was simple in the end. The "bad" widgets have some hidden overflow content in their page, which I don't want to include in the size. So if I take the size of the `<html>` tag instead of the body, then the visible size is reported, which is what I need.

It should be all good now, or at least much-much better.

gordonwoodhull:
Ah, that makes a lot of sense. It's often about measuring the right element! Thanks @gaborcsardi, and thanks @shaneporter for helping solve this!

gaborcsardi:
Another comment for the future. Broadly speaking it seems that there are two kinds of widgets:

1. Sets the widget height based on the width of the browser window (iframe for us) only, it does not care about the height of the browser window. This makes sense for a data table, or a radar chart.

2. Resizes the widget to the width and height of the browser window, with some padding. These widgets might look bad in a notebook, if the default height (400px currently) is too big or too small. It might make sense for the user to be able to manually resize the cell of the widget in this case. I don't think it is very important, and 400px will be probably OK almost always.

gordonwoodhull:
Great! Thanks again @gaborcsardi. 

Re: manual resizing, maybe later we can introduce jQuery UI resizing like we have for plots. (It's a bit harder because we probably can't detect hover on an iframe.) This is fine for now.

How does one currently control the sizing from mini.html? Is it about changing the size of the container? We should include it in the example (which will go into example notebooks):
https://rcloud.social/edit.html?notebook=4a2f823406f193aef51f04f86bc219b9

PR #2168, opened 2016-07-07
2151 remove binary assets from search index

### LINKED ISSUE #2167: Dont index binary assets in the search index
Refer discussion #2151

prateek05:
Closes #2167

### DIFF
--- a/rcloud.support/R/notebook.protection.R
+++ b/rcloud.support/R/notebook.protection.R
@@ -45,7 +45,7 @@ rcloud.set.notebook.cryptgroup <- function(notebookid, groupid, modify=TRUE) { #
     ## Remove notebook from SOLR Caching notebook was successfully assigned a crypto group
     if(is.notebook.encrypted(notebookid)) resp <- .solr.delete.doc(notebookid)
     ## If group change is set back to Public update the solr index
-    if(!is.notebook.encrypted(notebookid)) resp <- update.solr(rcloud.get.notebook(notebookid),rcloud.notebook.star.count(notebookid))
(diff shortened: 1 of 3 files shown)


prateek05:
@gordonwoodhull This should work.

gordonwoodhull:
Thanks @prateek05! Looks good.

gordonwoodhull:
I noticed one other small issue. 

The original call to `update.solr` did the check to see if SOLR is enabled outside the call. The new calls don't do this, so they are failing on non-search-enabled setups. Any objections to just exiting the function if SOLR is not enabled? I can make the change if this is OK.

prateek05:
The line from the original does exist. 
https://github.com/att/rcloud/blob/release/1.6/rcloud.support/R/search.R#L37 
Let me check

gordonwoodhull:
Yeah, but it's failing hard with `stop` rather than just returning. I think it should be okay to just skip it but I'm wary of messing with this code.

prateek05:
I think i traced it down to the deletion of notebook. Ideally the check should be moved to `.solr.post` to safeguard. Let me put a check there that should safeguard from all higher level calls

gordonwoodhull:
I noticed it on creating a notebook - `rcloud.set.notebook.visibility` calls `update.solr` which stopped.

prateek05:
Right and also `.solr.delete.doc`. 
ba5b4df the condition check is in the lower `.solr.post` that way its safeguarding any solr related functions to skip the process if the config does not have an entry for solr url

gordonwoodhull:
Great, rebased and merged (minor changes). Thanks @prateek05!

Issue #2150, opened 2016-06-29
discover page show times


gordonwoodhull:
While the spelled-out dates that we have are kind of nice, like Wed Jun 29 2016, it would be helpful to show the time, at least for recent notebooks. 

Maybe time-only for today, time and date without year for the past week, and date and year beyond that?

shaneporter:
Perhaps a nicer alternative would be something like:

'5 minutes ago'
'12 hours ago'
'3 days ago'

shaneporter:
That kind of thing...and so on.

gordonwoodhull:
Sure why not? That's sort of the github flavor.

I guess we could use [timeago](http://timeago.yarp.com/). Apparently [stackexchange does, or did](http://stackoverflow.com/a/3177900/676195).

shaneporter:
If the cool kids are using it...

gordonwoodhull:
There's this one too, using custom tags: https://github.com/github/time-elements

Judging by the tags, this might be what Github is using.

shaneporter:
Cool. And it looks like it works on Edge. 
So that's a relief.

shaneporter:
PR #2170.

Prateek032:
After editing Notebook, discover page fails to display the time in **Firefox**  browser.

Discover page before editing Notebook

![image](https://cloud.githubusercontent.com/assets/10941328/19031897/5b8ca410-8974-11e6-888f-fa37429af3d8.png)


and Discover page after renaming the Notebook

![image](https://cloud.githubusercontent.com/assets/10941328/19031930/a1d05c6e-8974-11e6-8b59-3d35bd16064b.png)

gordonwoodhull:
Thanks @Prateek032, please open a new issue for this.

PR #2170, opened 2016-07-08
Display relative dates/times on Discover

shaneporter:
I had intended to write 'Show times on Discover page', but actually, this PR doesn't do that, but I think that's fine, since our discussion on #2150 veered away from showing times to showing something a little more akin to Stack Exchange.

You can see the full date/time by hovering over the 'ago' text. It seems to use your locale, so I see

> 7 July 2016, 14:15 BST

On high traffic sites such as Stack Exchange and Twitter, this works nicely because it's easier to read '8 minutes ago' than a long string of text. On RCloud, however, perhaps we'll see longer periods of time such as 'a day ago' a lot more, so the time element is rather lost (except when hovering.) I'm not quite sure how much of an issue this is, but it rather moves away from the original intention.

In terms of JS, I adopted your last suggestion: that of a [custom element](https://github.com/github/time-elements) and its necessary [CustomElements polyfill](https://github.com/webcomponents/webcomponentsjs). Currently, it's only used on the discover page, so I kept it in that area, pushing the dependencies onto the discover require.

(Custom elements (directives) were one of the things I liked most about AngularJS, so it might be nice to use more of them in future.)


htdocs/lib/js/time-element.js
```diff
@@ -0,0 +1,616 @@
+(function() {
+  'use strict';
+
+  // Shout out to https://github.com/basecamp/local_time/blob/master/app/assets/javascripts/local_time.js.coffee
```
gordonwoodhull: I'd like to acknowledge the copyright a little more formally, e.g. a comment 
"This file derived from the Basecamp Local Time library
Copyright 2015 Javan Makhmali, Basecamp
MIT license"

gordonwoodhull:
Ahhh, this is nice, so much more friendly and calm. 

I'll apply the copyright notice change now.

PR #2172, opened 2016-07-08
Bump rcloud.support version to 1.6

prateek05:
Update rcloud.support package version to 1.6


gordonwoodhull:
Thanks @prateek05 - indeed we have not consistently done this. I'll add it to the release steps wiki too!

PR #2173, opened 2016-07-08
Cancelling from 'new asset' doesn't throw an error

shaneporter:
Because you can't use `trim()` on null.


gordonwoodhull:
Looks like this was deleted prematurely. Let's not trim any nulls!

I'm going to apply this manually.

shaneporter:
OK.

gordonwoodhull:
Surprising that you can view a commit on a closed PR with a deleted branch, but there seems to be no way to actually access the commit. Where did it go?

shaneporter:
I had a [shedload](http://www.oxforddictionaries.com/definition/english/shedload) of branches from 1.6 dev. I kept some, but this may have accidentally bit the dust. Sorry about that.

Issue #2078, opened 2016-06-15
paste or drop thumbnail


gordonwoodhull:
It's actually not very hard and it should excite people to populate the discovery page with their cool work.

But it's too late to be adding features to RCloud 1.6 right now

We should have a button in the asset panel (next to tabs) to the effect of ":heavy_plus_sign::camera:"

It brings up a small dialog where you can paste an image from the clipboard or drop a png onto the dialog. It previews the image, and if you press "Add Thumb" it adds/replaces the thumb.png asset.

Like what gmail and stackoverflow allow you to do. We already work with Blobs and data url PNGs and the clipboard and drag-and-drop and so on, so this is proven.

http://stackoverflow.com/questions/490908/paste-an-image-from-clipboard-using-javascript

gordonwoodhull:
The dialog is because
- we don't want paste normally to do this, so it's good to have an explicit target
- likewise drag and drop usually means "drop this asset with its current name" and there are other reasons why you might want image assets
- a preview is helpful because it often takes several tries to get a good screenshot, and that's a fair amount of data to write to the notebook.

shaneporter:
This duplicates the 'File Upload' functionality, but specifically for the thumb.

I imagine the next (much larger) iteration in this area is to have an integrated upload/assets panel. 

When 'Arrange panels by size' is checked, the 'File Upload' and 'Assets' panels are conspicuously separate.

shaneporter:
@gordonwoodhull I guess the file extension for this asset _must_ be .png? In which case, it should give UI feedback if it's not the correct file extension. (But I will show instructions on the dialog where necessary.)

gordonwoodhull:
Implemented in #2175

PR #2175, opened 2016-07-14
Paste or drop thumbnail dialog

shaneporter:
@gordonwoodhull - I'm submitting this PR now to give you a preview of where things are with the 'paste or drop thumbnail'.

Currently, paste is not supported, and although an asset is created/overwritten when the dialog's 'Update' button is clicked, it takes the original asset name, and doesn't write to `thumb.png`. I am using `RCloud.upload_with_alerts`. The file can't be renamed on the JS side. Is there something I can use to say 'write this to `thumb.png`'?

Other than that, it should be functional - I'm looking for some feedback on the UI really. Is it what you were expecting? 

If you could comment on the asset rename and UI, it'd be much appreciated. Thanks.


gordonwoodhull:
Yes, this is exactly what I was thinking - thank you! I love the little icon indicating that thumb.png is a special asset, and the hint there is nice too.

Hmmph, that's annoying that `upload_with_alerts` won't take a filename. It looks like it shouldn't be too hard to add an option at the the `RCloud.upload_assets` level, say `options.filenames`, an array that must be the same length as `options.files`. It may require adding a second parameter to the `operator` callback for `RCloud.utils.promise_sequence`, passing the index much like `Array.map` does.

I hadn't thought about showing the old and new thumbnails - I was imagining it would just show the current one or the one you've just added. Could make the dialog a bit simpler and smaller, but I don't have a strong opinion.

shaneporter:
I introduced the idea of the current thumb based on your statement in #2078:

> a preview is helpful because it often takes several tries to get a good screenshot, and that's a fair amount of data to write to the notebook

I thought that if you have a thumb already, it's good to compare what you're thinking about using against what you had before. Perhaps I've over-engineered this one, but the current one isn't shown if you don't have a thumb.

gordonwoodhull:
Yeah, I hadn't thought about comparing and it's a good idea but maybe it just feels a little more complicated this way, and I don't completely understand why. A couple guesses:

- When you're first adding a thumb, the thing being added is on the left. But when you replace, the current action is on the right.
- To determine what will happen, you have to read the text. Now I've got two icons in this dialog, does that mean I get two thumbs? Or should I click on one of them to choose? No, the one on the right is new and the one on the left is old. It takes thought and reading, which allows things down.
- There's a feeling of equality between the two thumbs 

A possible way to make it really obvious what's going on without needing to explain: if there's an old thumbnail, (temporarily?) make it the active asset so that it's showing in the background. Make sure the dialog doesn't overlap the asset pane. When the dialog opens, it shows the old thumbnail and you can drag or paste something there to replace. The one in the background in the asset pane stays the same until you click Replace and close the dialog.

This makes it obvious what is old and what is new, that nothing has changed until I click OK Replace, that the effect of the dialog is to change the asset which is showing underneath. 

I'm probably overthinking this. UI is just interesting to me. :blush: Let's roll with this for now. We can improve it later if we want to.

gordonwoodhull:
It's super redundant but I think the dialog should also have a file picker input. Sometimes it's easier just to go through the Open dialog than to drag or copy. And this is also friendlier for the less computer savvy.

shaneporter:
UI is interesting to me too; I will give it some thought.

Agree about the file picker.

As for the code changes on the R side, is that something that you can do, or should somebody else pick that up? Not sure how you're fixed at the moment. Thanks!

gordonwoodhull:
I don't think there are any code changes on the R side. All of the functions I mentioned are  in js/upload.js and js/utils.js. (But yes if we ever need server side changes I'm up for it.)

shaneporter:
Oh ok, I'll take a look. Thanks.

htdocs/js/ui/scratchpad.js
```diff
@@ -55,6 +55,14 @@ RCloud.UI.scratchpad = (function() {
                 RCloud.UI.thumb_dialog.init();
 
                 $('#update-thumb').click(function() {
+
+                    // select the thumb in the assets:
+                    var thumb = shell.notebook.model.get_asset('thumb.png');
+
+                    if(thumb) {
+                        thumb.controller.select();
```
gordonwoodhull: This doesn't.. quite.. work, right? If you want to be a hero, you could figure out why this helps and do something better:

https://github.com/att/rcloud/blob/develop/htdocs/js/notebook/asset_view.js#L62-L64

shaneporter: So you don't see the thumb when you click on the button? _Seems_ to work for me.

Ah - I see! (Or I don't see.) I'll take a look.

shaneporter:
I'm hopeful that this PR demonstrates the use of a specific dialog to update the notebook's thumb(.png) asset.

Features: 

- Image can be browsed, dragged or pasted when the 'thumb dialog' is visible;
- Verification prevents an invalid file type being used;
- Visual feedback occurs if an invalid file type is used (try using a .csv, for example);
- automatic selection of existing thumbnail, if any, when the dialog is shown.

Commit 10c790f fixes the 'where's my thumb?' issue that we were seeing. I have made some changes to the CSS, as well as removing the position bashing.

I noticed that the binary assets were shown differently to a text asset. Images, for example, are truncated, and you had to click the 'link' link to see the image in its full glory. The change I've made introduces scrollbars for binary assets, where necessary, so there's parity between the asset types. I also added a little 'external' (well, it's actually a 'share') icon to the 'link' so there's a hint that it'll open in a new window (or download.)

When updating a thumb, it does take quite a while for things to refresh, and although I was able to see the same issue by using the old drag and drop an asset method, the delay troubles me somewhat. Incidentally, we've been having [connectivity issues here in the UK today](https://www.theguardian.com/technology/2016/jul/21/bt-broadband-plusnet-outage-users-offline), so things may not be as bad usually.

gordonwoodhull:
This is awesome. Nice animation touches - we should do more of that.

I added "or Paste" and simplified the instructions slightly. Rebased and merged.

Thanks @shaneporter!

PR #2181, opened 2016-07-15
Updated SmokeTest.js 

Prateek032:


gordonwoodhull:
Thanks @Prateek032 - looks like a lot of the problem is proper use of promises, then. 

They are confusing beasts, to be sure.

PR #2182, opened 2016-07-15
Welcome back cell delete button, we missed you

shaneporter:
The people wanted a delete button at the cell level, so here it is.


gordonwoodhull:

PR #2185, opened 2016-07-15
Discover opens in a new page

shaneporter:
Discover opens in a new page.


gordonwoodhull:
ah, that was even easier than i thought. thanks @shaneporter!

PR #2188, opened 2016-07-18
Join cell button hidden for first cell on load

shaneporter:
The `check_buttons` function correctly shows/hides the 'join' button. This is called when a new cell is added, cell is deleted, cells are rearranged etc.

Code is then run that shows/hides the button based on its display_flags.  These are undefined, so it was then showing the button, regardless.

This show/hide operation is only done if the display_flags are not undefined.

### DIFF
--- a/htdocs/js/ui/cell_commands.js
+++ b/htdocs/js/ui/cell_commands.js
@@ -27,10 +27,13 @@ RCloud.UI.cell_commands = (function() {
                         commands_.map[cmd.key].disable();
                     else
                         commands_.map[cmd.key].enable();
-                    if(!_.every(cmd.display_flags, checkf))
-                        commands_.map[cmd.key].control.hide();
-                    else
-                        commands_.map[cmd.key].control.show();
+
+                    if(cmd.display_flags) {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Whoa, it's always weird to look back at old code you've written and try to figure it out as if someone else wrote it. I'm like, "above_between? huh? oh, that's not *so* bad..."

But it was bad. The problem here is that the flags system works, but the above_between system was working at cross purposes. Knowing me, I probably started replacing some old way of enabling and hiding controls with the flags system, thought that "between" was too complicated and didn't fit, and gave up and special-cased it.

`_.every(undefined, ...)` and `_.every([], ...)` are always true, and I don't want to break the contract that 

```js
                // a command will be enabled iff all of its enable_flags are true
                // a command will be shown iff all of its display_flags are true
```

So the right solution is to define a flag for this, and remove all the special cases for betweenness. Join is shown only if `!first`.

gordonwoodhull:
@shaneporter, I guess a lot of this junk is my mess to clean up. Sorry to throw you into the middle of this!

Issue #1706, opened 2015-10-08
Image asset goes blank when switched back to it from csv asset


sanketd11:
Steps:
- Upload any image and a csv asset
- Click on csv asset 
- Switch to image asset

The image asset goes blank. Image reappears if we click on any div from right side.

gordonwoodhull:
@anatoliyg, please test. You touched it, you bought it. :grin:

anatoliyg:
ok. must be z-index

PR #2190, opened 2016-07-20
images assets are shown correctly in assets

shaneporter:
So I've made some changes, and cannot reproduce #1706, or the 'where's my thumb' issue we were seeing on PR #2175,

However, I get the feeling I'm missing something because I see no reason why the `object` element should have a `position: absolute` property/value applied to it. With that in mind, there is no need to switch between `static` and `absolute` in `asset_view.js`.

### DIFF
--- a/htdocs/js/notebook/asset_view.js
+++ b/htdocs/js/notebook/asset_view.js
@@ -59,9 +59,6 @@ Notebook.Asset.create_html_view = function(asset_model)
     anchor.click(function() {
         if(!asset_model.active())
             asset_model.controller.select();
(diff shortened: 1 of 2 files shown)


shaneporter:
though the `thumb.png` asset does take an unnervingly long time to display first time.

shaneporter:
Ah ok - I see an issue - the 'link' is not shown where it should be.

shaneporter:
Closing. Will reopen another PR when I have a solution.

gordonwoodhull:
That's probably why the image was position absolute, just so the link could be relative. But it could just as well be the parent or grandparent with position absolute or relative.

Issue #1734, opened 2015-11-02
Add a way to clear the Session pane


CourseraStudent58:
The contents of the Session window is very helpful.
But as it gets long and you have to scroll a lot, it would be helpful if it could be cleared between runs.

s-u:
You can always remove cells that you deem unnecessary. If you start new work, you just create a new notebook.

CourseraStudent58:
Perhaps I am not using the correct terminology.  I don't mean a new RCloud Notebook session.  Rather, I am referring to the "Session" tab that appears in the lower right section of the RCloud UI.  When a user Rund their code, log info appears in this Session tab.  I normally have debugging sequences where I Edit-Run-Edit-Run...and all the while, the log info is appending inside the Session tab.  I'm suggesting that I'd like to "clear" that Session tab between each Edit-Run iteration.

s-u:
Ah, ok, got it, yes, makes sense - feature request to clear the Session pane.

gordonwoodhull:
I often want a Reset Session button - usually to clear JavaScripts but also to clear R variables. 

I wonder if we should combine these features, and further whether it would be nice to have an option to have the Run All button also reset the session. Because right now you aren't sure if your notebook really works if you press Run All. Simple ex: `library(X)` is in a cell after package `X` is used.

prateek05:
+1 on clearing R for RunAll

gordonwoodhull:
There are two requests here:

 - [ ] an option (default enabled) to reset the R session when pressing (or invoking with what's-that-hotkey) Run All. Resetting the session always clears the session pane.
 - [ ] a little button in the header of the session pane to clear the content like the new notebook button. (This would benefit from #463 but could be implemented independently.)

gordonwoodhull:
The button could be 
![image](https://cloud.githubusercontent.com/assets/1366709/15562271/dd48fd2e-22cb-11e6-9ce3-79359a968412.jpeg)
only white on black

CourseraStudent58:
I like first option [an option (default enabled) to reset the R session when pressing (or invoking with what's-that-hotkey) Run All.].  Reason: less clutter and user can override option for backwards compatibility behavior.  Plus, the proposed button for the second option is ginormous! ;-)

shaneporter:
@gordonwoodhull I like the button; but I'm guessing you'd prefer it if it were a little smaller? :smile:

gordonwoodhull:
[...]

PR #2198, opened 2016-07-22
Clear session button

shaneporter:
Part of #1734, this adds an 'eraser' button to the session panel heading area and clear the panel contents when clicked.


gordonwoodhull:
Thanks @shaneporter!

PR #2202, opened 2016-07-26
Highlight imported notebook(s)

shaneporter:
This code supports highlighting of imported notebooks either by 'Import notebook from file' or 'Import External Notebooks'.

For multiple notebooks, each one is scrolled to and highlighted in order. Unsure if the `effect` timings - currently at 1.5s, are a little long, but I don't _think_ they are.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -845,6 +845,17 @@ var editor = function () {
         ui_utils.scroll_into_view($tree_.parent(), 50, 100, $(node.element));
     }
 
+    function highlight_imported(node) {
+        var p = node.parent;
+        while(p.sort_order===ordering.NOTEBOOK) {
+            $tree_.tree('openNode', p);
+            p = p.parent;
+        }
+        ui_utils.scroll_into_view($tree_.parent(), 150, 150, $(node.element), function() {
+            $(node.element).closest('.jqtree_common').effect('highlight', { color: '#fd0' }, 3000);
+        });
+    }
+
     function select_node(node) {
         $tree_.tree('selectNode', node);
         scroll_into_view(node);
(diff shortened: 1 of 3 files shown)


htdocs/editor_tab.js
```diff
@@ -1477,6 +1494,19 @@ var editor = function () {
...
+            var nodes = _.map(_.isArray(notebooks) ? notebooks : [notebooks], function(notebook) {
+                return $tree_.tree('getNodeById', node_id('interests', username_, notebook.id));
+            });
+
+            // get promises:
+            nodes.map(function(node) {
+                return highlight_imported(node);
+            }).reduce(function(cur, next) {
+                return cur.then(next);
+            }, Promise.resolve()).then(function() {});
```
gordonwoodhull: That is an extremely clever map-reduce promise! I didn't know you could do that, but I guess I shouldn't be surprised, since functional programming always has a way. This is what [`promise_sequence`](https://github.com/att/rcloud/blob/develop/htdocs/js/utils.js#L11-L20) does.

gordonwoodhull:
The timing is good, but maybe they should animate simultaneously? It does feel slow in the Import External Notebooks case. 

I guess you're concerned about notebooks that land all over the tree, hmm.

shaneporter:
Yes, because the scroll implies they may be off screen, so highlighting might not be seen.

htdocs/editor_tab.js
shaneporter: :blush: I didn't know about [`promise_sequence`](https://github.com/att/rcloud/blob/develop/htdocs/js/utils.js#L11-L20).

shaneporter:
The users will let us know if they're not happy :smile:

gordonwoodhull:
This is going to be very helpful, and it's good to have a careful and patient animation when introducing a feature.

gordonwoodhull:
Thanks @shaneporter. Rebased and merged.

Issue #2192, opened 2016-07-20
search again hotkey


gordonwoodhull:
Cmd/ctrl-G (and F3 on Windows/Linux) goes to the next occurrence of the search term (or previous, with shift-cmd/ctrl-G / shift-F3). 

If the find dialog is not open, opens it.

shaneporter:
Currently, if 'next' is clicked, the match index advances. If the find dialog is closed, and reopened, that match index resets to the first match.

So for this:

> If the find dialog is not open, opens it.

Should 'go to next' start at the first item (or the second?) and
Should 'go to last' go to the first, or the last?

My feeling is that go to next should start at the beginning, and go to last should go to the last one.

gordonwoodhull:
That was my thought initially, but at least in Chrome, the find command opens continues where it left off. 

Arguably that's different because their Find control closes when you click on anything. 

It looks like the standard behavior (looking at TextEdit and LibreOffice) is to *continue from the cursor position*. That's pretty smart. Not simple in our case, but we've already done part of that for populating Find from the current selection.

gordonwoodhull:
That's also nice because it can be the same behavior for the case where the Find dialog has lost focus (and thus we've lost the results) as for when the Find dialog has been closed.

shaneporter:
Hmm... I'm more used to shift + f3 being used for 'go to previous' than 'go to last match' (Windows/Linux.)

gordonwoodhull:
Think I wasn't careful with my words there, that's what I meant. Editing now.

shaneporter:
PR #2205.

PR #2205, opened 2016-07-26
Next/previous search result hopping

shaneporter:
Should implement the required functionality for #2192.


gordonwoodhull:
This is a good start. But the command isn't enabled when a cell is active. 

I'd also expect some attempt to continue either from where the last find let off, or from where the cursor or cell selection is, when the find dialog has lost focus. But maybe we should put that part off.

gordonwoodhull:
This is actually a really complex little feature, so maybe we should put it off. Search "again" isn't quite as useful if it starts over from the beginning each time it's invoked.

shaneporter:
Hi @gordonwoodhull - please could you take a look at this when you have a moment.

gordonwoodhull:
I don't see the keys mapped for cells, so it's hard to test this. (Was there a file left out of the commits?) The code looks good otherwise.

Also, it doesn't set focus to the find "dialog" if it's already open, so nothing happens in that case.

shaneporter:
Whilst looking at this, I found that initiating a search with 'shift-f3' or 'ctrl-shift-g' goes to the first match, rather than the looking back from the cursor.

gordonwoodhull:
Right, I can invoke the command only outside of the cells (as a global hotkey), not from within a cell where it's most useful. So the ace hotkeys need to be modified for cells.

I was figuring this was the reason it was going to the first match rather than continuing from the current cursor position: because there is no current cursor position.

gordonwoodhull:
Looks like what happens is that Ace already has a different find next and find previous, which we don't want. (It looks for the currently *selected* text.)

But it's bound to different keys on Mac and Windows, so it's letting you take the key on Windows:

```js
JSON.stringify(widget.commands.commands.findnext, null, 2)
{
  "name": "findnext",
  "bindKey": {
    "win": "Ctrl-K",
    "mac": "Command-G"
  },
  "readOnly": true
}
```

Adding the following lines to the cell view's `create_edit_widget()` fixes it (in 56de411bf87):
```js
        ace_widget_.commands.removeCommand('findnext');
        ace_widget_.commands.removeCommand('findprevious');
```

gordonwoodhull:
Works great!

shaneporter:
Thanks for the explanation @gordonwoodhull!

PR #2207, opened 2016-07-27
Asset images no longer overlap other panels

shaneporter:
This seems to fix the Firefox issue of asset panel images overlapping other panels. Tested in Chrome and there is no apparent issue from the changes.

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -196,7 +196,8 @@
       <div id="scratchpad-editor-wrapper">
         <div id="scratchpad-editor" class="widget-vsize">
         </div>
-        <div id="scratchpad-binary"></div>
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Put a div around it! Ain't that always the way.

shaneporter:
Haha, though I don't usually add a div to solve a problem. In this case, it seemed like the simplest way, and that old weird position switching has gone, and you can scroll (albeit only in the y direction) the asset, so it's not all bad :smile:

gordonwoodhull:
Too often, that seems to be the way to fix stuff, especially when involving scrolling. But I'm still kind of a n00b after a few years doing this full-time. Someday I'd love to learn how browsers actually work.

Issue #2219, opened 2016-08-03
htmlwidgets are slow


gordonwoodhull:
Moving discussion from att/rcloud.rcap#98

@gaborcsardi wrote:

> @gordonwoodhull OK, here is a plan. We'll only send scripts (and data?) over once. On the client side, we store them in a dictionary. We use hashing to see if we need to send a script over or not.
> 
> For this we need to parse the HTML to find the tags that I want to extract and hash. Can I depend on the xml2 package for the HTML parsing? Or does rcloud have a standard way to parse HTML on the server side?
> 
> Btw. it would be even better if I could serve the scripts from the server, over standard HTTP. Is there a standard rcloud way to do this?
> 

@gaborcsardi, htmlwidgets uses htmltools. So I think you can use the htmltools API to read/write the dependencies, and you won't need to parse HTML. In an experimental version of rcloud.shiny, I used this to rewrite the URLs.

Serving over HTTP is possible, but it's a little bit more complicated than you might guess. For Shiny, the session **is** the server, so it can use the URL namespace however it wants, including creating a bunch of temporary directories. 

We can't do that, since there are many sessions running at the same base URL. It's a namespace issue. Each session could stomp on the URL namespace of all the others, and the session may not be running on the server you're connected to (because of load balancing). 

So we offer the following two mechanisms instead.

 * If the asset is in the `inst/www/` directory of a package, it can be fetched with `shared.R/package/<asset>`

 * Or you can use `proxy.R` to serve the scripts over HTTP from your session, but you'll need to run your own HTTP server from R.

`proxy.R/<user>/<session>[:<port>]/[<path>]`
https://github.com/att/rcloud/blob/develop/htdocs/proxy.R

Yes that is `:port` inside the path portion of the URL. Usage example using `rcloud.session.info()` to get the session id: https://github.com/att/rcloud.shiny/blob/feature/call-serviceApp/R/rcloud.shiny.R#L45-L48

The (untested) direction I think rcloud.shiny is going is: 
 1. allow Shiny to spawn its own server, but don't block on it
 2. poll that server via a timeout from the client
 3. create an iframe and point to the proxy.R URL to allow Shiny to serve its own `/`
 4. now the Shiny server owns its own URL namespace and can do whatever it wants

That's a possible approach for htmlwidgets as well. In the non-viewer mode, it launches an HTTP server and enjoys its own URL namespace to serve assets.

gordonwoodhull:
[...]

PR #2231, opened 2016-08-11
htmlwidgets improvements

gaborcsardi:
Addresses #2219.

* We do not use the `viewer` option any more, but define our own `print.htmlwidget` method. This allows more flexible HTML code generation.
* JS "dependencies" are not embedded into the HTML now
* The function that handles widgets with the `suppress_viewer` class is not needed any more. We handle them automatically, by defining a `print.suppress_viewer` method.
* `shared.R` can serve files from `inst/htmlwidgets`. This is how the dependencies are served.
* `shared.R` allows caching for 1 hour.

### DIFF
--- a/htdocs/shared.R
+++ b/htdocs/shared.R
@@ -30,5 +30,5 @@ run <- function(url, query, body, headers) {
   f <- file(fn, "rb")
   r <- readBin(f, raw(), s)
   close(f)
-  list(r, mime::guess_type(base))
+  list(r, mime::guess_type(base), "Cache-Control: max-age=3600")
 }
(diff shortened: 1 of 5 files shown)


gaborcsardi:
Some more comments.
* Yes, this makes all widgets **a lot** faster. :) Especially when the JS files are already cached.
* I am using the `htmlwidgets` and `htmltools` packages, but do not mention these packages in `DESCRIPTION`. This is a failure in `R CMD check`, but for us it is OK, because if you want to print an HTML widget, then you'll have to have these installed. Do you want me to add an explicit dependency? Maybe a soft `Suggests`?

gordonwoodhull:
Wow, that's fast!! I have rebased and merged this, thanks!

I'll trust your judgment on the dependency, as I don't know too much about R packages and @s-u is on vacation. Should I add `htmlwidgets, htmltools` to  the `Suggests` list in `rcloud.support/DESCRIPTION`?

PR #2235, opened 2016-08-15
Fix 2234

### LINKED ISSUE #2234: adding/changing comments fail in SOLR call
Follow-up to #2225 - once the visibility check was fixed, I discovered a couple of other bugs.

@prateek05, it looks like the format for the structure we're saving comments in SOLR has changed. Both adding a new comment and changing an existing comment fail with this error:

```
Error in solr.res$response$docs[[1]] : subscript out of bounds
```

The error appears in the session pane. 

Please check.

prateek05:
Fixes #2234

### DIFF
--- a/rcloud.support/R/search.R
+++ b/rcloud.support/R/search.R
@@ -206,7 +206,7 @@ stitch.search.result <- function(splitted, type,k) {
   # Create reponse
   res <- list()
   res$id <- id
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Phew, that is much more readable and maintainable than it was. 

Tests good, too. 

I did discover one new issue but I don't think it's bad enough to hold up the point release.

PR #2245, opened 2016-08-24
Changing cell language subsequently sets focus to the cell's content

shaneporter:
With a simple call to `cell_view.edit_source(true)`.


gordonwoodhull:

PR #2246, opened 2016-08-30
Asset blur on 'escape'

shaneporter:
Press `esc` whilst focus is on an asset's contents to blur the ACE widget.


gordonwoodhull:

Issue #2220, opened 2016-08-04
Revert Notebook should be removed from the Keyboard shortcuts list


Prateek032:
For read only notebooks (when we load any other user's Notebook) **`ctrl e`** is listed in keyboard shortcut window.

gordonwoodhull:
I suggest adopting a more active flags-based design instead of just modes, "lifting" the flags support in the cell commands extension into `RCloud.extension` itself, and thus having the ability to enable/disable commands based on flag or !flag.

https://github.com/att/rcloud/blob/develop/htdocs/js/ui/cell_commands.js#L4-L37

gordonwoodhull:
Specifically, Revert should be disabled when it's another user's notebook (and arguably when you're on the current version, although it's a no-op in that case). The logic that shows/hides the Revert button in the navbar seems to get it right, if a little complicated.

https://github.com/att/rcloud/blob/develop/htdocs/js/ui/configure_readonly.js#L8-L21

Maybe the navbar should be using flags too.

shaneporter:
PR #2271.

PR #2247, opened 2016-08-30
This prevents the 'revert notebook' shortcut being active for read only notebooks

shaneporter:
The original issue, #2220, had a good suggestion from @gordonwoodhull about using rcloud extension type flags based 'enabling' of shortcuts.

I have fixed the issue, so 'revert' should not be shown/enabled for read only notebooks, but would prefer to update the code to use a flags based approach, as per the suggestion, in another PR.

### DIFF
--- a/htdocs/js/ui/init.js
+++ b/htdocs/js/ui/init.js
@@ -179,6 +179,7 @@ RCloud.UI.init = function() {
             ]
         },
         on_page: ['edit'],
+        modes: ['writeable'],
         action: function() {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
The fix is incorrect because it disables Revert even when looking at a historical version of one's own notebook. So the command is completely inaccessible (because one can't revert the current version anyway).

I think we've run into this complication before. It's not the same logic as readonly. I'll clarify in #2220.

PR #2250, opened 2016-08-30
Search/replace NaN bug fixed

### LINKED ISSUE #2248: Search: clicking previous/next for not found content results in NaN being displayed
Steps to reproduce:

1. Open find or find/replace.
2. Enter text that will not be found, verify that `0 of 0` is shown in the search field.

image

3. Click the previous or next button and observe the 'NaN of 0' that is shown:

image

shaneporter:
This PR fixes #2248 where 'NaN' was being shown incorrectly for the 'current match' in the search results.

### DIFF
--- a/htdocs/js/ui/find_replace.js
+++ b/htdocs/js/ui/find_replace.js
@@ -17,7 +17,7 @@ RCloud.UI.find_replace = (function() {
 
         // matches_
         find_match_[matches_.length === 0 ? 'addClass' : 'removeClass']('no-matches');
-        show_match_details(matches_.length === 0 ? '0' : '1', matches_.length);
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I got bogged down while reviewing this - it changes half of the occurrences of `active_match_ !== undefined` to `matches_.length != 0` but not the other half, which makes me think that something is poorly specified here.

I noticed that if you enter a search string which does not exist, and then click Replace, you get an error on the console. This error is not caused by this PR - it's an old bug, and seems to be caused by the same condition. So maybe we should use `matches_.length != 0` everywhere? 

Could you take another look at this and see if we can make it more consistent and robust?

shaneporter:
Of course.

gordonwoodhull:

Issue #2213, opened 2016-07-29
Firefox: Match count is not getting displayed for find/replace


amithrb:
For editable cell, the match count is not getting displayed for find/replace

Steps to Reproduce:
- Create new notebook
- Add new cell
- Add content to newly created cell
- Save it
- Invoke find/replace option
- Search for content
- Close find/replace option using ESC or Close option.
- Once again invoke find/replace option

Below is the screenshot
![capture](https://cloud.githubusercontent.com/assets/19602717/17244604/5d22f8cc-55a0-11e6-80b5-5e9ef28fab5f.PNG)

gordonwoodhull:
This is related to #2214.

For me, a recently-used menu is left behind when the dialog closes. If I click away, it invokes #2214, highlighting the results. If I invoke find while the recently-used menu is shown, then find appears with no match count shown.

@amithrb, is this what you see as well?

gordonwoodhull:
menu left behind floating in space
![image](https://cloud.githubusercontent.com/assets/1366709/17365481/5c940402-5954-11e6-836d-c0308e975a59.png)

shaneporter:
Suggested fix in #2249.

gordonwoodhull:
fixed by #2251

Issue #2215, opened 2016-07-29
Firefox: On clicking ESC, unable to close invoked find/replace feature


amithrb:
Unable to close invoked find/replace feature, on clicking ESC button.
This issue is not reproducible on rcloud-5 and rcloud social but able to reproduce on local instance.

Steps to Reproduce:   
- Add contents to the cell
- Make cell non editable
- Invoke find/replace option
- Close find/replace option using ESC or Close option
- Click anywhere on the screen, contents gets highlighted
- Once again invoke find/replace option
- Now, try to close invoked find/replace feature using ESC

gordonwoodhull:
Thanks @amithrb. In the fifth step, do you mean "click on the cell" instead of "click anywhere on the screen"? I'm not sure what you mean by "contents get highlighted".

Prateek032:
@gordonwoodhull , After closing  **Find/Replace**, click on cell/ any where on the screen. 
Contents are getting highlighted 

![image](https://cloud.githubusercontent.com/assets/10941328/17247557/a1390388-55b1-11e6-9f68-02c9a1b8eb2a.png)

gordonwoodhull:
Thanks @Prateek032. I see it now.

gordonwoodhull:
I see the highlight problem (#2214) but was not able to repro ESC not closing the dialog. It makes sense that it could be caused by the recently-used menu, so I'm going to trust that #2251 fixes this.

Issue #2214, opened 2016-07-29
Firefox: Previously searched content gets highlighted 


amithrb:
On clicking anywhere on the screen, previously searched content gets highlighted.

Steps to Reproduce:
- Add contents to the cell
- Make cell non editable 
- Invoke find/replace option
- Close find/replace option using ESC or Close option
- Click anywhere on the screen, contents gets highlighted

gordonwoodhull:
As noted in #2213, I think this is related to the recently-used menu that is left behind when the dialog closes.

shaneporter:
Suggested fix in #2249.

gordonwoodhull:
Seems this didn't have to do with the recents menu, as #2251 didn't fix it.

shaneporter:
I _could_ reproduce this issue, but am no longer able to do so using the 'steps to reproduce' at the top.

amithrb:
@shaneporter , even am not able to reproduce this issue, working fine for above mentioned steps.

PR #2251, opened 2016-08-31
Fixes for various find/replace bugs

shaneporter:
This PR seems to fix issues #2213, #2214, #2215.

- The autocomplete now disappears on Firefox when the dialog closes.
- The change_interval_ was accidentally global, and although its code checked the old/new value in the find field, the matches weren't shown (#2213) when the dialog was shown and the value hadn't changed. (it doesn't change if you close/reopen of course), so `generate_matches` is called manually.


gordonwoodhull:
I can still repro @2214 highlights reappearing on click, but I think this fixes the other two. Rebasing and merging. 

Thanks @shaneporter!

PR #2252, opened 2016-08-31
Find/replace matches always shown, regardless of current selection

shaneporter:
`z-index` change for find match means that the 'match' background will be higher than that of any current selection, and result in the orange (current match) or yellow (match) being shown.


gordonwoodhull:

Issue #2204, opened 2016-07-26
cursor should select current find result when find closed


gordonwoodhull:
It's useful to use Find to place the cursor where you want to edit. At least when the notebook is editable, we should probably activate the cell of the current result and select the current result.

This is particularly useful in combination with Find Again #2192, because often you want to do something more complicated at each site, editing each result.

Find is a marvelously complicated little feature! It's remarkable how few applications get every feature right.

shaneporter:
Suggested resolution in #2254.

PR #2254, opened 2016-09-06
Cursor selects current find result when find closed

shaneporter:
This is an implementation for #2204, where the current match is selected when find is closed.


gordonwoodhull:
This works great, and should be really helpful.

However, there's a small glitch: if the current result is not found, then `current_match` is undefined and it becomes impossible to close the find dialog.

shaneporter:
Please try again @gordonwoodhull. Thanks.

gordonwoodhull:

PR #2255, opened 2016-09-06
solr url config item should be commented

shaneporter:
Under the 'Optional Functionality' > 'Search' section of `install.md`, solr installation is inferred as optional.

In that case, `solr.url` should be commented out in the `rcloud.conf.samp` file; otherwise it will cause solr related issues if solr is not installed.


gordonwoodhull:
I agree. Thanks!

gordonwoodhull:
merged

PR #2260, opened 2016-09-08
fixed NaN current match after replace all, find previous

shaneporter:
Fix for issue found by @Prateek032, where NaN is shown for current match after 'replace all' and 'find previous' are clicked.

### DIFF
--- a/htdocs/js/ui/find_replace.js
+++ b/htdocs/js/ui/find_replace.js
@@ -34,7 +34,7 @@ RCloud.UI.find_replace = (function() {
     };
 
     function matches_exist() {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Note: other failing condition is https://github.com/att/rcloud/issues/2248#issuecomment-245498438

gordonwoodhull:
We still have the weird behavior where old matches remain highlighted and replaceable after they've already been replaced with something that doesn't match. Guess I'll open a new issue for that.

But better not to display NaN in that case!

gordonwoodhull:

PR #2261, opened 2016-09-08
Modifier click + target shortcut support

shaneporter:
Intended to be a 'documentation style' shortcut, the new `click_keys` shortcut property enables the following type of shortcut:

'Shift click on the play button'

The object structure is like this:

````javascript
click_keys: {
    target: 'Play button',
    win_mac: [
        ['shift']
    ]
}
````

The `win_mac` property is optional, and follows the pattern of existing shortcuts, which has one of `win`, `mac` or `win_mac` properties set.

I have added an example into `init.js` (see below). 

One aspect I'm not entirely happy with is the display of the shortcut in the shortcut dialog:

image

The 'shift' + 'key(s)' and 'target' parts are enclosed in a `kbd` element, but I think this can be better.

This won't be a big change of course.

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -447,7 +447,11 @@
             <td>
               <% _.each(shortcut.keys, function(key_combo, index) { %>
                 <%if(index > 0) {%> / <%}%>
-                <kbd><%=key_combo%></kbd>
+                <%if(_.isObject(key_combo)) { %>
+                  <kbd><%=key_combo.keys%> click</kbd> <span>on the</span> <span class="target"><%=key_combo.target%></span>
+                <% } else { %>
+                  <kbd><%=key_combo%></kbd>
+                <% } %>
               <% }); %>
             </td>
             <td>
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -135,8 +135,8 @@
 
 <script type="text/html" id="help-snippet">
   <div id="help-wrapper" class="panel-body widget-vsize">
-    <a href="#" id="show-shortcuts" title="Show Keyboard Shortcuts">
-      <i class="icon-keyboard"></i>Keyboard
+    <a href="#" id="show-shortcuts" title="Show Shortcuts">
+      <i class="icon-keyboard"></i>Shortcuts
     </a>
     <div id="help-header">
       <form id="help-form" action="#">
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by gordonwoodhull** [comment]
This is great - I'm going ahead and adding the obscure modify-clicks for cell selection now. I hadn't understood the obvious fact that these are alternate methods for accessing commands which belong in existing categories, rather than a separate category, so thanks for that.

I agree the formatting is a little awkward - maybe part of it is that the target isn't a key, and not always a button. Since "click" is already a verb, it could almost do without formatting or prepositional phrase, so:

<code><kbd>shift click</kbd> play button</code>  Run from this cell on

<code><kbd>ctrl click</kbd> cell title</code> Toggle cell selection

I'm not completely happy with that either, just throwing it out there. Let's populate it a bit more and see how it shakes out.

**Comment by shaneporter** [comment]
I think you're right - it's an example of needing more, ahem, examples to figure something out.

**Comment by gordonwoodhull** [comment]
Okay, I tried adding the shortcuts for cell selection but there is more than one place which assumes there will be keyboard shortcuts and not just click commands.

I'm starting with this one:

```js
{
        category: 'Cell Management',
        id: 'select_cell',
        description: 'Select individual cell',
        click_keys: {
            target: 'Cell title',
            win_mac: [
                []
            ]
        }
    }
```

(The modifiers are intentionally empty because this is just a click.)

This works if I also add `keys`, but in this case and others, there are no keys.

First problem is in the line 

```js
            if (shortcut.keys.hasOwnProperty('win_mac')) {
```
https://github.com/att/rcloud/blob/36f6b6de07fd0cffc8614c82b581903c9deb2f2c/htdocs/js/ui/shortcut_manager.js#L56

but even if I put that in an if-block, the shortcut still doesn't show up in the help dialog. So I figure I should pass this back to you.

shaneporter:
I was under the impression that the new items also used key modifiers. If they don't, it's debatable whether they should appear on the '_Keyboard_ Shortcuts' dialog. :confused:

gordonwoodhull:
Haha I guess we could rename it "Shortcuts", that's kind of catchy. 

OTOH, this is likely the only exception, since there isn't really anything else you can click on that's not obvious. Will anyone notice the inconsistency, or worry where they can find help for a command that doesn't use any keys?

shaneporter:
"Shortcuts" sounds good to me. I reckon that should cover it.

shaneporter:
I've made some changes that allows this to be done:

````javascript
    {
        category: 'Cell Management',
        id: 'select_cell',
        description: 'Select individual cell',
        click_keys: {
            target: 'Cell title',
        }
    }
````

note that a `win_mac`, `win` or `mac` property isn't required.

gordonwoodhull:
we'll definitely want to terse up these descriptions somehow.

image

still, it's great to have it documented!

gordonwoodhull:
added a few in the linked commits. it's a start!

Issue #2008, opened 2016-06-02
File gets downloaded while uploading to the assets


Prateek032:
When we try to upload document(.docx) file to the assets, the file itself is getting downloaded.
Note: this issue observed only on Chrome browser (Version 51.0.2704.79 )

gordonwoodhull:
Thanks @Prateek032.

To be precise, this happens whenever the docx file is shown. 

@shaneporter, we may be trying to do something too tricky here. We attempt to use the archaic `<object>` tag in order to embed any content of any type. And then we blacklist pdf in particular because we know it doesn't work. Should we continue blacklisting, or should we whitelist? Or should we use a more specific tag for the types that work, like `<img>`? 

Please investigate, sometime (but not a priority for this release).

https://github.com/att/rcloud/blob/develop/htdocs/js/ui/scratchpad.js#L188-L191

shaneporter:
PDFs don't work because the server's `Content-Type` response header value is `text/plain` rather than `application/pdf`.

Take a look at this plnkr I created, requesting an online PDF and a server responding with the correct `Content-Type` header:

http://plnkr.co/edit/ABmj5Qsh26dkhilZBHsB?p=preview

Perhaps you already knew all that. But I've tested the plnkr in Chrome and Firefox and I see the PDF, and the only difference appears to be the `Content-Type` header.

What types of binary content does the preview area claim to support? Is it images? Thus far, it seems a little confusing since it _seems_ as though PDFs _could_ work, yet they're blacklisted. Everything else is given a chance, but this bug results from Chrome's 'incorrect' handling of `docx` files (and `pptx' files, amongst others, too.)

shaneporter:
I plan to whitelist the following file types:

- bmp
- jpg/jpeg
- png
- gif
- svg?

When the server code is changed to respond with the appropriate `Content-Type` response header, PDF can be added to the supported file types.

For non-supported file types, an appropriate message can be shown; the user can always get a hold of the file by clicking on the 'link' link.

shaneporter:
PR #2262.

PR #2262, opened 2016-09-09
Asset file types handled more effectively

shaneporter:
Initially a fix for #2008, this PR removes the catch-all `object` tag and implements `img` tag handling for image types (bmp/jpg/jpeg/png/gif).

[PDF support seems possible](http://plnkr.co/edit/ABmj5Qsh26dkhilZBHsB?p=preview), so I have added some specific code for PDF handling. This can be uncommented when the server returns the appropriate response header.

### DIFF
--- a/htdocs/js/ui/scratchpad.js
+++ b/htdocs/js/ui/scratchpad.js
@@ -205,7 +205,6 @@ RCloud.UI.scratchpad = (function() {
                 // ArrayBuffer, binary content: display object
                 $('#scratchpad-editor').hide();
 
-                // PDF seems not to be supported properly by browsers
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Awesome, I added `pdf = "application/pdf"` to notebook.R and verified it fixes PDF display when separately opened. Now let's see if that works in the asset view too...

gordonwoodhull:
Yay! The size is wrong but PDF works now. Tested in Chrome, Firefox, Safari on OSX.

It's been broken for a long time, good to know it was just the mime type. Thanks @shaneporter!

shaneporter:
Great!

gordonwoodhull:

PR #2266, opened 2016-09-12
Right at end/Left at start of ACE widget moves to next/previous cell

shaneporter:
This permits, using right and left cursor keys, the ability to traverse, where appropriate, to the next and previous cell.

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -350,6 +350,10 @@ function create_cell_html_view(language, cell_model) {
         ui_utils.install_common_ace_key_bindings(ace_widget_, function() {
             return language;
         });
+
+        var left_handler = ace_widget_.commands.commandKeyBinding[0].left,
+            right_handler = ace_widget_.commands.commandKeyBinding[1].right;
+
         ace_widget_.commands.addCommands([{
             name: 'executeCell',
             bindKey: {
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
Thanks @shaneporter. I think the up and down keys would be more intuitive - why the change here?

shaneporter:
Oh right! Yes, at some point in the past I thought of using 'left' and 'right' for this, but yes, you've mentioned 'up' and 'down' keys everywhere. Sorry; I don't know why I became fixated with left/right!

So, it's just a case of swapping 'left' for 'up', and 'right' for 'down'? I can do that.

gordonwoodhull:
Cool! I think it's a tiny bit simpler because you only have to check line and not column.

shaneporter:
Yeah, you're right. I'll make the change. Good to be simpler.

shaneporter:
So, the rules should be:

- Go to previous cell: if there is a previous cell and __up__ arrow is pressed and you're on the first line
- Go to next cell: if there is a next cell and __down__ arrow is pressed and you're on the last line.

shaneporter:
So, I was going to remove the left/right handling, but I think it compliments the (albeit more intuitive) up/down handling.

Try it - I think the left/right should stay, but happy to remove the left/right if you have strong objections.

gordonwoodhull:
You know me, I like having lots of ways of doing things. The left/right is awesome too.

Arguably up/down should go to the same character instead of the beginning (or end of line if shorter) but we can think about that later.

shaneporter:
Good point about up/down.

PR #2267, opened 2016-09-12
Remove SASS reference

shaneporter:
The change from [this comment](https://github.com/att/rcloud/pull/1868#issuecomment-202477028) means that the SASS reference is not required.


gordonwoodhull:
Good call - thanks! Rebased and merged.

PR #2268, opened 2016-09-13
Asset view PDFs have more room to breathe

shaneporter:
CSS and a conditional class for the object/embed height.


gordonwoodhull:
Great!

PR #2269, opened 2016-09-13
Setting for clear R session

shaneporter:
Seems as though I missed something from #1734. Here is the setting - and I _think_ that it's now over to @gordonwoodhull for the actual clearing of the R Session.


gordonwoodhull:
Thanks for the nudge, connected it up and closed #1734!

PR #2271, opened 2016-09-15
Revert notebook shortcut shown appropriately

shaneporter:
Introduced an `is_active` function to the shortcut. This uses the same logic as the revert toolbar item to only show the shortcut when appropriate.


gordonwoodhull:

PR #2279, opened 2016-09-17
Cross cell traversal preserves column position

shaneporter:
This maintains the column position when traversing cells with up/down. No special handling for 'at end of line if new line has fewer characters' - ACE puts cursor at end of line automatically.


gordonwoodhull:

PR #2280, opened 2016-09-17
up/down cell shortcuts documented and no more 'spaced out'

shaneporter:
- addition of up/down cell shortcuts
- realised that 'spaced out' issue related to incorrect shortcut key structure.


gordonwoodhull:

PR #2287, opened 2016-09-30
Find next/previous, replace shortcuts

shaneporter:
Try as I might to get mousetrap (shortcut library) to recognise (shift-)f3 inside the search form, but it didn't seem to work.

Therefore, I've piggy-backed onto the existing 'find_next_on_enter' (renamed with 1042c4a) to trap f3/shift f3.

The replace shortcut should only be active within the search form (added `element_scope` with 5d99024).

Find next/previous now documented. Replace shortcut documented.

### DIFF
--- a/htdocs/js/ui/find_replace.js
+++ b/htdocs/js/ui/find_replace.js
@@ -177,7 +177,7 @@ RCloud.UI.find_replace = (function() {
             replace_cycle_.push('find-close');
 
             function find_next_on_keycode(e) {
-                if(e.keyCode===$.ui.keyCode.ENTER || (!ui_utils.is_a_mac() && e.keyCode === 114)) {
+                if(e.keyCode===$.ui.keyCode.ENTER || (!ui_utils.is_a_mac() && e.keyCode === 114 /* f3 */)) {
                     e.preventDefault();
                     e.stopPropagation();
                     if(e.shiftKey)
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Since these are modal, I think it would be helpful to give them their own category in the Shortcuts Help.

Also, ctrl/cmd-R is a very dangerous hotkey. (I wasn't thinking, just like with ctrl/cmd-N.) To make it somewhat safer, may I suggest <kbd>alt r</kbd>? And why not do <kbd>alt a</kbd> for replace all, while we're at it.

gordonwoodhull:
Oh, I see, it's to hoist those functions to the outer scope. 

Maybe they should be moved out instead. Seems like I left a lot of junk inside `toggle_find_replace` for now reason - probably just because I was rushing the thing, thinking find & replace was a simple feature.

shaneporter:
@gordonwoodhull I've made some changes, except the `toggle_find_replace` refactoring.

I'm thinking that it might be easier to merge the refactoring into the develop branch once PR #2205 has been merged.

gordonwoodhull:
Sure, that makes sense. Always tricky to deal with multiple branches/PRs on the same code.

Which one would you like me to merge first?

shaneporter:
I guess #2205, so I can then do the refactoring as part of this PR.

gordonwoodhull:
I did the refactoring, rebased and merged. Thanks @shaneporter!

PR #2289, opened 2016-10-02
All of the find 'input' accepts focus

shaneporter:
Focus should now be set when any part of the find 'input' field is clicked.


gordonwoodhull:
Thanks @shaneporter. You should also set `cursor: text` for the find details, to indicate what a click will do. Right now it changes back and forth between cursors as you mouse left and right over the input.

shaneporter:
Good catch! Updated with f1acece.

gordonwoodhull:
Weird, you can still detect where the invisible `#match-status` is by clicking multiple times - on the text input area, it will show the previous values. But who cares about that.

PR #2298, opened 2016-10-07
Editable title behaving on Firefox

shaneporter:
This PR should resolve issues relating to Firefox and editable titles (notebook title and tree titles).

Implemented explicit handlers for HOME, END and RIGHT.

### DIFF
--- a/htdocs/js/ui_utils.js
+++ b/htdocs/js/ui_utils.js
@@ -434,6 +434,9 @@ ui_utils.editable = function(elem$, command) {
         range.setEnd(text_node, position);
         selectRange(range);
     };
+    function getCaretPosition() {
+        return window.getSelection().getRangeAt(0);
+    };
     function options() {
         return elem$.data('__editable');
     }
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @shaneporter!

This solves most of the problems, however there are a couple more (at least on OSX):
 - right arrow still cancels in the notebook title as the first action, when the full content is still selected. It does work in the notebook tree, comments, and asset names.
 - alt-right arrow (equivalent of ctrl-arrow on winux) should move one word right but it still exhibits cancel behavior at the end of the notebook title (but works everywhere else).

shaneporter:
So the 'good' news was that I was able to reproduce and fix the issues without a Mac. Please take another look. Thanks :)

gordonwoodhull:
Oh fun, implementing word-left and word-right by hand. 

I had to correct `metaKey` to `altKey` in 3b6440be8f6. `metaKey` is <kbd>cmd</kbd>

Seems like this disabled cmd-left and cmd-right for beginning and end of line, in some places but not others (?!?) But the original bug is fixed, so I think it's time to put this down and move on.

gordonwoodhull:

PR #2307, opened 2016-10-17
Session pane performance (thanks to infinity.js fork)

shaneporter:
This seems to have made a difference. The [fork of Airbnb's infinity.js](https://github.com/elliotcw/infinity) seems to have done the trick.

I added a comment at the top of the added `infinity.js` file.

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -254,7 +254,7 @@
 </script>
 
 <script type="text/html" id="session-info-snippet">
-  <div class="panel-body tight" id="session-info-panel">
+  <div class="panel-body tight scrollable" id="session-info-panel">
     <div id="session-info" class="widget-vsize"></div>
   </div>
 </script>
--- a/htdocs/js/ui/session_pane.js
+++ b/htdocs/js/ui/session_pane.js
@@ -1,5 +1,6 @@
 RCloud.UI.session_pane = {
     error_dest_: null,
+    list_view_: null,
     clear_session_: null,
     allow_clear: true,
     body: function() {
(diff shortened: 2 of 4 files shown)


gordonwoodhull:
Boy is it tedious testing this. Takes more than minute to return in Chrome, slightly better in Firefox but not enough.

gordonwoodhull:
It does seem to help the lag in scrolling (once it finally shows at all).

Sorry but maybe it's not the right solution after all. I guess I was hoping that infinity.js wouldn't require us to create all the elements until they're needed, and I didn't understand that it's just swapping the elements in. So the lag is still pretty severe when printing. Note that R comes back in less than 10 seconds, so any lag after that is DOM manipulation.

Also, the session pane does not always grow to show the text. This is probably because infinity.js makes something asynchronous, so the size is not known when `right_panel.collapse` does a resize. 

Finally, on Firefox, the output does not show at all if the first output is huge.

gordonwoodhull:
Seems like to use this strategy, we'd have to find another library that only generates elements when they're needed (and preferably removes them when they're not visible. 

It's not documented and they call it experimental, but [the infinite scrolling feature of iScroll](http://iscrolljs.com/#infinite-scrolling) seems to work this way.

It looks pretty simple - if you check out the source of [the demo](http://lab.cubiq.org/iscroll5/demos/infinite/), it just has an undocumented `dataset` callback that retrieves a range of data. We don't need any of the ajax stuff, but we would need to store an array of output strings and supply it in response to `dataset`.

It looks like this library is also actively maintained.

gordonwoodhull:
Guess I'll close this now. Sorry for the wild goose chase!

PR #2309, opened 2016-10-18
Can now paste into find field

shaneporter:
I've tested this on Firefox and Chrome. Pasting is permitted, and the `change` event binding picks up on the pasted change.


gordonwoodhull:
Always nice when you can fix a bug by simplifying the code! 

PR #2310, opened 2016-10-20
Pull and replace

shaneporter:
Invoke 'Pull and Replace from notebook' on the advanced menu. Dialog presents three source options: file, URL and Notebook ID.

The file import functionality was already present in 'Import notebook from file', so rather than duplicating this, I moved it to `util.js` in the `Notebook.read_from_file` function. This takes an options parameter, so when things go wrong, or the import is successful, calling code can be notified. This new code is now used in the 'import/export' code, as well as the 'pull and replace' code.

The URL code parses the URL and looks for the notebook parameter value. It then calls the ID code, which imports the notebook from ID.

If anything goes wrong, an error message is shown inline. In addition, there are checks so that the current notebook cannot be pulled into itself. I guess this wouldn't have been an issue as such, but it doesn't really make sense.

If the import is successful, `rcloud.set_notebook_property` is called to retain the details of the import. This doesn't apply to files (since the file has to be uploaded again), but saves time for ID and URL pulls.

When the notebook is being pulled, I adopted a 'modify dialog' approach, hiding 'close' elements and changing the 'Pull' button so that it is effectively disabled and its text is modified to 'Pulling'. I'm aware of the 'blind/dialog' approach used elsewhere, but didn't want to have a dialog over a dialog UI situation.


htdocs/edit.html
```diff
@@ -154,6 +154,41 @@
   </div>
 </script>
 
+<div id="pull-changes-dialog" class="modal fade in">
+  <div class="modal-dialog">
+    <div class="modal-content" style="background-color: rgba(255, 255, 255, 1.0)">
+        <div class="modal-header" style="padding-left:20px!important;padding-right:20px!important">
+            <button type="button" class="close" data-dismiss="modal" aria-hidden="true">x</button>
+            <h3 class="modal-title">Pull and Replacae from notebook</h3>
```
gordonwoodhull: typo

gordonwoodhull:
Great work, @shaneporter. A few notes.

IMO `URL` should be the first option, and the default. This will be the most common option.

The notebook history is not being refreshed. If you Undo after pulling and replacing, you go to the version before the one where pull & replace was invoked. I.e. you undo the replace, but instead you go to the version before that. There is also a version missing from the history in the notebook tree.

It looks like you may need to call `update_pulled_by()` before `get_input().val(...)` because it will attempt to set whichever control is active from the last time the dialog was invoked, instead of the last option used for this notebook.

So for example if I replace from file in one notebook, then go to another notebook which has been pulled & replaced by URL, and invoke Pull & Replace , I see the following error:
```
InvalidStateError: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.
```
and the dialog fails to open in that case. The code correctly wouldn't try to restore a file input, but since it's between states, it tries to stuff the URL into the file input.

htdocs/edit.html
shaneporter: That's my classical education.

shaneporter:
I'm now resetting the dialog state, and doing the `update_pulled_by` with an additional (optional) `value` parameter. This seems to have resolved the file input issue.

cff299f seems to have resolved the notebook tree and history state after a pull.

gordonwoodhull:
Works great now. Thanks @shaneporter!

Rebased and merged.

PR #2311, opened 2016-10-20
Cell-based export

shaneporter:
This adds a setting, 'Export only selected cells'.

Choosing 'Export notebook to file' or 'Export notebook as R Source file' will export only those cells that are selected.

If no cells are selected, I thought it should default to exporting everything. That's somewhat in disagreement to 'Export only selected'; what it means is "if there's something selected, export only that, otherwise, selected everything". Perhaps something a little more pithy is required for the UI though :)


gordonwoodhull:
Looks great. It has the usual problem of the default coming out undefined thus false, until the option is set. I've rebased and merged and I'll open a separate issue for that. Thanks @shaneporter!

PR #2315, opened 2016-10-25
Export only selected uses correct default value when setting is NULL

shaneporter:


gordonwoodhull:

PR #2325, opened 2016-10-28
Fixes #2324

### LINKED ISSUE #2324: Cookie Domain do not infer until explicitly set
The issue is with systems which do not know their external FQDN. (AWS, Docker, Any virtualized container , etc)

We currently explicitly set cookie.domain obtained from the what the OS reports. 
https://github.com/att/rcloud/blob/develop/rcloud.support/R/setup.R#L190
`if (!nzConf("cookie.domain")) setConf("cookie.domain", getConf("host"))`

We further go on to restrict for a must have to provide the domain from the backend. 
See: login.R login_successful.R goodbye.R

Since the frontend does a good job of inferring the domain from the access URL it would expose probably a flaw with anyone setting up a reverse proxy of any domain  in between the client and the backend. 

Should this be an added/optional feature. If explicitly set, the backend provides the domain if not set fall back to letting the client decide from the access url ?

For further info see #2317

prateek05:
Fixes #2324


prateek05:
@gordonwoodhull @s-u 
Would this be fine ? To make if optional ?
Original issue #2324

prateek05:
@s-u thoughts ?

s-u:
There is one real issue (empty list creates a cookie entry of just "=") but also it's an attempt to make it more generic but still copy/paste everything. So IMHO we should do either a fully generic replacement (add a function to construct cookies in rcloud.support) or do only minimal update to support not setting the domain (which is needed by Docker and places where deployed behind a proxy with unknown host).

gordonwoodhull:
@prateek05, did @s-u's comment make sense to you? Are you willing to give this another shot?

prateek05:
Yes. I ll give it another go this week.

Issue #2321, opened 2016-10-28
Pull and replace from notebook: Pull option is still being displayed as Pulling even after reopening it


Prateek032:
Steps to reproduce:
- Copy any existing notebook ID
- open any other user's Notebook (read only mode)
- then open Pull and replace from notebook
- paste the copied notebook id in the field
- Click on pull

In session div it throws the error (as expected), but the `Pull` option is still displayed as `Pulling`
![image](https://cloud.githubusercontent.com/assets/10941328/19801675/da8b449e-9d1d-11e6-9e7c-95cff6e69d84.png)

after closing the modal window(using `Esc` key). If we again open the pull and replace from notebook option, it still display `Pull` as `Pulling`

![image](https://cloud.githubusercontent.com/assets/10941328/19801776/4501f71e-9d1e-11e6-8ff0-5bed4f339bc9.png)

gordonwoodhull:
I'm fixing the error handling now - a promise was dropped. I'm surprised that the error reported is `404 Not Found` - maybe GitHub is deliberately vague for security reasons.

@shaneporter, the action item here is to disable Pull&Replace when the current notebook is read-only.

shaneporter:
PR #2327.

PR #2327, opened 2016-10-31
Cannot Pull and Replace Notebook when current notebook is readonly

shaneporter:
An action point from #2321. My interpretation of 

> disable Pull&Replace when the current notebook is read-only.

Is to recreate how the 'Open in GitHub' menu item works: test, alert, return.


gordonwoodhull:
Well, I meant literally disable, like we do for Publish Notebook. 

Took me a while to find the example though.

PR #2328, opened 2016-10-31
Fixes Issue #2323

### LINKED ISSUE #2323: Infinite login loop when no real authentication backend exists
In deployments with file-based storage and no real authentication backend. 
The login loop does not rewrite the new tokens 

Issue is largely with here:
https://github.com/att/rcloud/blob/develop/htdocs/login.R#L61
`if (!is.null(getConf("exec.auth")) && !isTRUE(cookies$user == usr)) ` 
The logic needs to be further expanded to account for an invalid token. 

More details see Issue #2256

prateek05:
Fixes #2323


prateek05:
@s-u  Would this be fine ?

PR #2329, opened 2016-10-31
Pull and Replace error message improvements

shaneporter:
- Error message is more appropriate in the event of an invalid URL.
- Error message in case of GitHub 404 has been improved ('The notebook could not be found')


gordonwoodhull:

PR #2333, opened 2016-11-02
Disabled menu items have a noop action

shaneporter:
The main change has been in the menu code. If an item is disabled the item's function is assigned to a `enabled_action` property, and a noop is assigned to the item's action. On enabling an item, this `enabled_action` property is tested and reassigned to the item's `action` property. If it doesn't exist, then the default `action` is used.

In addition, I thought it would be good UX to inform the user _why_ that item is disabled.

So I added an optional `disabled_reason` property to items (you might want to improve the 'Publish Notebook' item, since it doesn't say _why_). This is then set as the item's `a` title attribute.

### DIFF
--- a/htdocs/js/ui/menus.js
+++ b/htdocs/js/ui/menus.js
@@ -58,6 +58,21 @@ RCloud.UI.menu = (function() {
                     var item = extension_.get(menu_item);
                     if(!item || !item.$li)
                         throw new Error('menu disable fail on ' + menu_item);
+
+                    if(!enable) {
+                        item.enabled_action = item.action;
+                        item.action = function() {}    
+                        if(item.disabled_reason) {
+                            item.$li.find('a').attr('title', item.disabled_reason);
+                        }                
+                    } else {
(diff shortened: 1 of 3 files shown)


htdocs/js/ui/menus.js
```diff
@@ -58,6 +58,21 @@ RCloud.UI.menu = (function() {
...
+                            item.$li.find('a').attr('title', item.disabled_reason);
+                        }                
+                    } else {
+                        if(item.enabled_action) {
+                            item.action = item.enabled_action;
+                            delete item.enabled_action;
+                        }
+                        item.$li.find('a').removeAttr('title');
+                    }
+
```
gordonwoodhull: Whoops, if it gets disabled twice (e.g. two readonly notebooks opened in a row), then the `action` is completely lost. Replacing with a disabled flag which gets checked before `.action()` is invoked.

shaneporter: Ah! Thanks!

Issue #2302, opened 2016-10-11
"real star" or bookmark feature for own notebooks


s-u:
Users have a nice feature to "star" other's notebook they find noteworthy. However, there is no such thing for your own notebooks, because by definition all own notebooks are starred. I would find a feature useful where I can tag some of my notebook as extra noteworthy - sort of real star as opposed to the automatic star.

Perhaps an alternative would be to not star all notebooks by default, but I wonder what other implications that would have ...

s-u:
PS: another interesting thought would be to have the automatic star go away with time - the thought being that whatever I start working on is relevant in the short term, but only things I tagged should stay relevant in the long term...

PR #2335, opened 2016-11-02
API: create can handle a form submission now

gaborcsardi:
Plus:
* do a 302 redirect after the notebook is created
* show an error page if user is not logged in

### DIFF
--- a/htdocs/api.R
+++ b/htdocs/api.R
@@ -10,12 +10,13 @@ cookies <- function(a) {
 }
 
 create.notebook <- function(c, caps, body) {
-  body <- rawToChar(body)
+  if (is.raw(body)) body <- rawToChar(body)
   content <- fromJSON(body)
   new.nb <- RSclient::RS.eval.qap(c, as.call(list(caps$rcloud$create_notebook, content)))
   if (!isTRUE(new.nb$ok)) stop(paste("failed to create new notebook",toString(new.nb),sep='\n'))
 
-  list(paste0('{"redirect":"/edit.html?notebook=', new.nb$content$id, '"}'), "application/json")
+  headers <- paste0("Location: /edit.html?notebook=", new.nb$content$id)
+  list("", "", headers, 302)
 }
 
 run <- function(url, query, body, headers) {
@@ -43,7 +44,7 @@ run <- function(url, query, body, headers) {
     anonymous <- FALSE
     init.cap <- caps$rcloud$session_init
     if (is.null(init.cap))
-      return(list("Unauthorized", "text/html", character(), 401))
+      return(list(unauthorized.page, "text/html", character(), 401))
     RSclient::RS.eval.qap(c, as.call(list(init.cap, cookies$user, cookies$token)))
     verb <- pex[1L]
     et <- paste("Error executing", verb)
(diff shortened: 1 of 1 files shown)


htdocs/api.R
```diff
@@ -55,3 +56,25 @@ run <- function(url, query, body, headers) {
...
+    <link rel="stylesheet" type="text/css" href="/css/rcloud.css"/>
+  </head>
+  <body id="goodbye">
+    <div class="navbar navbar-inverse navbar-fixed-top">
+      <div class="navbar-header">
+        <a class="navbar-brand" href="/login.R">RCloud</a>
+      </div>
+    </div>
+    <div class="container" id="main-div" style="margin-top: 100px;">
+      <p>To create an RCLoud notebook you need to be logged in, but
```
shaneporter: RCloud

gaborcsardi: Thanks!

gordonwoodhull:
I'm going to leave this open until I see it work - I guess the problems with RStudio Export happen earlier in the chain, but I figure it's easier just to keep the branch/PR open.

We also need to star and show the notebook, something I neglected in my initial implementation. By default notebooks are not visible to other users, and we star them so that they'll show up in what passes for current notebooks in the tree. The star design is somewhat disputable (#2302) but that's what we've got for now.

Usually we do this [from the javascript side](https://github.com/att/rcloud/blob/7d562735613c2bb2a1e7b7ce0720a390226cb748/htdocs/editor_tab.js#L1394-L1395) but they are simple APIs on the R side too.

gaborcsardi:
OK, I do the store-token-redirect-pickup loop now if the user is not logged in. Seems to work fine.

I'll do the staring in a minute.

gaborcsardi:
@gordonwoodhull I am not sure what `RSclient::RS.eval.qap(c, as.call(list(caps$rcloud$create_notebook, content)))` returns, but apparently I cannot use that to star the notebook.....

How do I star the new notebook?

gordonwoodhull:
`caps$rcloud` should contain all the authenticated OCAPS, [which are listed here](https://github.com/att/rcloud/blob/develop/rcloud.support/R/ocaps.R). I don't know what that `RSClient` stuff does TBH but I think if you repeat the pattern you should be able to call any of them.

gordonwoodhull:
Er, looks like there was some careless editing in ocaps.R, the stars function [are actually in the unauthenticated ocaps](https://github.com/att/rcloud/blob/35cf4469beed930f91fcada16966f2fdbac0e25c/rcloud.support/R/ocaps.R#L162), which the authenticated ocaps inherit.

gordonwoodhull:
Okay, I guess I'll take the create/show bits. 

It occurred to me that we should really have an `api.R/loggedin` boolean `GET` rather than pushing the data and stashing it. So I guess I will document `api.R` in the wiki with a big warning that it's subject to change.

gaborcsardi:
YEah, we don't have to advertise API imo, it is not a real API, anyway.

I am not sure how much `api.R/loggedin` would help, because I cannot query it without CORS.

gordonwoodhull:
Inevitably everything becomes an API over time. 😜 

That's a good point about same-origin problems. It's probably not good to rely on cookies for a REST API, and we'll need to pass some token as a parameter instead.

gaborcsardi:
Exactly. Another reason I would not call the current code an API just yet.

As I see it, the current form submission is a hack, and a proper REST api should use tokens. But that would make the RStudio addin somewhat inconvenient.

gaborcsardi:
Btw. in theory we could also read out the cookie from the local browser I guess. See e.g. https://gist.github.com/dacort/bd6a5116224c594b14db (I haven't tried this.)

gordonwoodhull:
Thanks @gaborcsardi, we'll leave these clues for later. 

Here is the anti-documentation (feel free to edit if inaccurate): 
https://github.com/att/rcloud/wiki/api.R

gaborcsardi:
OK, looks good to me. :)

gordonwoodhull:
Thanks @gaborcsardi, rebased and merged.

PR #2342, opened 2016-11-07
Pull and Replace fixes

shaneporter:
Some fixes for Pull and Replace to hopefully tidy things up a little in that area.

I noticed that the way I had toggled the 'Pull' button's enabled status was in conflict with the other dialog's action buttons, which do not get disabled based on user input. Rather they, and now the Pull and Replace 'Pull' button are always enabled, and errors are shown when the button is clicked.

This has resulted in a few deletions, making the code simpler.

### DIFF
--- a/htdocs/js/ui/pull_and_replace.js
+++ b/htdocs/js/ui/pull_and_replace.js
@@ -26,7 +26,9 @@ RCloud.UI.pull_and_replace = (function() {
                     // update pulled by method:
                     update_pulled_by(type, value);
                 }
-                else update_pulled_by('url');
+                else {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Doesn't look like we have much consistency in that department - Import from File has disabled Import until file selected and if description is not empty. but Import External Notebooks apparently does not prevalidate.

It's somewhat more friendly to check inputs early and only enable the submit button when there's a chance of success. But let's go ahead and fix these bugs and worry about the perfect interface whenever.

gordonwoodhull:

PR #2345, opened 2016-11-08
Pull and replace from same notebook prevented

shaneporter:
This PR tests, when pulling from ID, if that ID is the current gistname.


gordonwoodhull:
Thanks @shaneporter, rebased and merged.

PR #2363, opened 2016-11-22
Fix Solr Async Calls on updates

prateek05:
The latest curl seems to resolve the SSL bug when curl calls are forked

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+## RCloud 1.8
+
+### Improvements
+* Solr index update calls are non blocking (Requires R 3.3.x)
+
+
 ## RCloud 1.7
 
 In addition to the below changes to RCloud core, this release also supports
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Thanks @prateek05. Could we also specify the curl version where this is fixed, in `../DESCRIPTION`? (Was RCurl the culprit?)

prateek05:
No it was the underlying libcurl 7.19.7-52. With R 3.3.x on RHEL 6 we have to build libcurl independently from the system one (Does support 7.19.7). 
This is tested with libcurl/7.50.1 OpenSSL/1.0.1e zlib/1.2.8 and the original issue only affected https calls.

gordonwoodhull:
I see, thanks for clarifying. How about adding a note to the release notes (NEWS.md) saying that a newer libcurl is required, then? Or I suppose we could require R 3.3, but that's kind of harsh.

prateek05:
Right. The added dependency on curl greater than 7.19.7 comes with the dependency chain of R 3.3.x. This is only tied to systems like RHEL6. Should i mention it in NEWS ? 
Prateek

gordonwoodhull:
Yes, please start an **RCloud 1.8** section in NEWS and add a note there. 

Then I think we are good to merge. Thanks!

gordonwoodhull:
Thanks @prateek05!

PR #2368, opened 2016-11-30
Modified failing scripts

NaikPrateek:


Prateek032:
Updated following scripts
-  tests/SmokeTest.js
-  tests/Run.sh
- scripts/create-travis-configuration.sh

New scripts related to following features
- Pull and replace
- Flexdashboard.html
- Import Rmarkdown file
- html widget related scripts

gordonwoodhull:
I don't think this fixes the current errors but I'm merging it in case.

Issue #2371, opened 2016-12-06
view (shareable link) should trigger save first


gordonwoodhull:
any action that is affected by the state of the notebook should save the notebook.

Issue #1728, opened 2015-10-23
delete artifacts?


gordonwoodhull:
there is a lot of build that has to happen when someone installs rcloud from the source.

the only thing folks are avoiding by having the artifacts checked in, is installing npm in two directories.

couldn't we automate the install of npm and avoid this? or since we're only using uglify in order to validate, use some other linting tool that doesn't require npm?

gordonwoodhull:
(motivation is that merging or overwriting artifacts is extremely annoying to developers)

s-u:
+1

gordonwoodhull:
Proposed solution. @s-u, @prateek05, is this acceptable?

 1. make `npm` another prerequisite (currently we only require it if you're ["hacking on the code"](https://github.com/att/rcloud/blob/develop/doc/INSTALL.md#will-you-be-hacking-on-the-code-read-on) but since we require `R` and a few other things, seems reasonable enough to require node.
 2. [de](https://github.com/att/rcloud/blob/develop/htdocs/js/package.json)-[duplicate](https://github.com/att/rcloud/blob/develop/htdocs/lib/package.json) `package.json`, move it to root, and make it somewhat descriptive of RCloud (even though we still won't actually publish RCloud to npm, of course)
 2. do `npm install` in root directory when `bootstrapR.sh` is run
 3. add a dependency to a plain javascript parser like acorn or esprima, for validation with decent error messages
 4. validate before uglifying, pointing to root directory's `node_modules`
 5. consider disabling uglify.

s-u:
I'm somewhat hesitant making `nodejs` + `npm` a requirement for everyone since it's not directly available, e.g. on OS X. I think we should really keep the distinction of hacking on it - where it's ok to require a lot more and work off the bare git checkout - and installing it which doesn't need either. So if we just provide automated tar balls (like we do for packages on RForge.net) which contain the artifacts then I think it would be ok.

gordonwoodhull:
Ah - so you're suggesting distibuting through RForge rather than GitHub?

s-u:
I don't think GH has a way to build distribution tar balls (since you don't have a VM to install things on). I didn't check, though, if there is a way to "push" distribution tar balls that has been built elsewhere - if it does, then we could do that, otherwise either out RCloud website or RForge.net would be ok ...

gordonwoodhull:
Okay, now I've lost interest because this is stuff I don't now how to do.

s-u:
[...]

PR #2405, opened 2017-05-02
Add resolution of _htmlwidgets urls for users packages

useless5771:
this fixes att/rcloud.htmlwidgets#6

The problem here was that the additional '_htmlwidgets' urls were just supported for packages installed globally in rcloud. When user tried to use a package that got installed into their library directory, the '_htmlwidgets' urls were not resolved.


gordonwoodhull:
Fantastic! Five fixes!

I know it's extra work, and it doesn't matter much for small fixes like these, but in the future, please open a branch and a new PR for each fix. 

It's easier for me to review, and in case I need to selectively merge some fixes but not others, I won't leave some commits dangling that way.

useless5771:
Yes, I noticed that. I will raise pull requests from feature branches from now on.

useless5771:
@gordonwoodhull  please could you review this PR and merge the commits if they are fine?

(I have three bug-fix branches for which I would like to raise PRs but they are based on changesets in this pull request, so when I try to raise new pull request github includes these not-yet-merged changes in them as well.)

Thanks

gordonwoodhull:
Yes! I do apologize for the delay. Today is 90% RCloud for me, so I should have this merged by EOD if not sooner.

gordonwoodhull:
Ah, one more procedural request (sorry). Usually when submitting a PR, it's customary not to include the generated artifacts (in this case `*_bundle.*`), since those just add noise to the commits and make the merges slightly more difficult.

I wish we didn't have these checked in at all (#1728), but until we solve the tooling/distribution problem, please omit these from your commits. This is a best practice for most open source projects. 

Thanks!

useless5771:
Yes, I was surprised that these are not ignored, just assumed that since https://github.com/att/rcloud/blob/develop/doc/INSTALL.md suggests that node.js and npm, are somewhat optional then these are expected to be committed... (as minified and source get out of sync)

Ok, I will exclude these from my commits then.

gordonwoodhull:
Thanks... Yes, we're in a bit of a rut where we need them committed for every release, because we want people to be able to run RCloud without Node tools, but we don't want to commit them when we're between releases because it's a mess.

Once we develop a process for pushing built releases to GitHub, we can finally stop checking this junk in.

gordonwoodhull:
Hi @useless5771 - I was not able to verify the first fix on this branch, but the others look great, so I've merged everything except 355cba6.

The fix for #2371 is not guaranteed to work because the default action for the link opening a new tab runs in parallel with the save. But I appreciate that you did it this way because it's better than using javascript to follow a link. In practice this should be fine and of course it tests well. 

(I only noticed this because I'm paranoid about dropping promises after many bad experiences.)

I'll return to the original fix for `shared.R` when I can figure out how to repro multiple-user installs in a test environment. Hopefully merging the rest is enough to get you unstuck on your other PRs.

gordonwoodhull:
I put the single remaining commit on #2407

useless5771:
Thanks @gordonwoodhull , I sorted out the mess with our fork history and raised four pull requests based on current rcloud head so it should be quick to review/merge these.

PR #2407, opened 2017-05-08
Add resolution of _htmlwidgets urls for users packages

gordonwoodhull:
fixes att/rcloud.htmlwidgets#6

This PR replaces #2405. I merged the rest of it but this one I was unable to test and thoroughly review due to my test machine only supporting a single user.


gordonwoodhull:
@s-u, could you take a quick look at this since it's a big chunk of R code? I haven't been able to test it since I unfortunately need to run `devtools::install_from_github` in order to install the parcoords package, and that doesn't work due to some permissions problems on romo.

gordonwoodhull:
I couldn't change the target of this PR (it needs to go the release branch instead of develop).

So I'm closing this and replacing it with #2419

PR #2408, opened 2017-05-09
Fix #2369 by ensuring that recent notebooks list is always refreshed

### LINKED ISSUE #2369: if notebook fails to load, recent notebooks are not available
menu is empty.

useless5771:


gordonwoodhull:
Good solution - I had thought of just running this promise chain in parallel without relying on the success of load, but using `finally` is clearer and probably more correct. 

I'm just simplifying it slightly (sort of), to keep methods of the editor_tab returning promises directly.

PR #2409, opened 2017-05-09
Fix #2209 by handling errors during import of notebooks

### LINKED ISSUE #2209: Unable to import multiple external notebooks if one of the Notebook is invalid
While importing multiple notebooks if one of the notebook is encrypted then it fails to import the remaining notebooks.

useless5771:


gordonwoodhull:
Thanks @useless5771! Rebased and merged as 31ffb76825c6

PR #2410, opened 2017-05-09
Fix #2139 by exposing redirect url parameter on shell.do_load method

### LINKED ISSUE #2139: if loading notebook fails due to logout, reconnect should still send you to selected notebook
1. Login to second browser and do stuff, to invalidate token on first browser
2. Click on notebook in tree in first browser

Message: authentication required, reconnect?

On reconnection, sent to current notebook instead of the one that was clicked.

I hadn't seen this before because RCloud used to let me log in on two browsers. But now I am seeing it all the time.

useless5771:
Also fix fatal_dialog so it can actually be reused with different url, label and buttons set.

### DIFF
--- a/htdocs/js/rcloud_bundle.js
+++ b/htdocs/js/rcloud_bundle.js
@@ -782,11 +782,12 @@ ui_utils.make_url = function(page, opts) {
     return url;
 };
 
-ui_utils.relogin_uri = function() {
+ui_utils.relogin_uri = function(redirect_url) {
+  redirect_url = redirect_url || window.location.pathname + window.location.search;
     return window.location.protocol +
         '//' + window.location.host +
         '/login.R?redirect=' +
-        encodeURIComponent(window.location.pathname + window.location.search);
+        encodeURIComponent(redirect_url);
 };
 
 ui_utils.disconnection_error = function(msg, label) {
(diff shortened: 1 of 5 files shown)


useless5771:
bundle.js sneaked in, I will remove this.

gordonwoodhull:
No problem, I just rewrote history. (Glad this doesn't work in the real world.)

In the future, you could simply add another commit reverting the file, and then it's easy to squash. Or it's actually okay to do a force push on a pull request branch; github will just recalculate the PR. (A PR really just tracks the differences between two live branches.)

Great work - thanks for making `fatal_dialog` more reusable, and improving the interface around `do_load`. This is an extensive change and it looks like you tied up all the loose ends, but it's hard to tell. Let's see how it shakes out.

This will help a lot because it's easy to lose concentration when loading a new notebook and being forced to redirect. I bet there are a bunch of other places we find we want to use the improved redirects.

gordonwoodhull:
I don't know why, but this is causing error messages to be replaced "more". 

In particular, when the client code version does not match the server code version, we display a fatal error telling the user to force-reload. 

Great. But then we also shutdown the connection, which produces a lost connection error which we should not display because it will replace a specific error with a general one.

Generally I think we want to show the first error in the fatal dialog, because it will be the root cause and the most specific. If errors cascade they should go to the debug console but not change the fatal dialog. 

So I don't know why there was already this code:
```js
    else message_.text(message);
```

I'm guessing it was a last minute hack to get some particular case working. 

For some reason, I always find my self "fixing" the error handling right before a release, and I suspect I'm about to break some other error condition. Sigh. But yeah, now I'm removing the whole second else-block, as well as some misbegotten error "handling" in `rclient.js` which was reporting `[Object object]`. 😝

PR #2411, opened 2017-05-09
Disable upload button when there is nothing to upload, fixes #2390

### LINKED ISSUE #2390: upload button should be disabled until file selected
not like anyone would get confused and keep pressing that button and get mad when nothing happens (like i just did).

useless5771:


gordonwoodhull:
Rebased and merged. Thanks @useless5771! No more false button rage!

PR #2412, opened 2017-05-10
Fix #2334 by exposing hide/show buttons on selection bar

### LINKED ISSUE #2334: Close all cell output button on banner
Please provide a mechanism to close a cells output without being at the top of a cell.  Log cell outputs can make closing them painful.

Another option could be to right click in the cell to have the option to close it.

Yet another option would be for the workbook to remember the state of cells and keep the output in the closed or open state.

I feel that the first option is the easiest to implement since it is literally a loop calling your close function on each cell or open function on each cell when invoked.

useless5771:
These buttons result in invoking the same action as the toggle results button from the cell-bar so the border around the toggle results button appears when applicable.

### DIFF
--- a/htdocs/js/notebook/notebook_model.js
+++ b/htdocs/js/notebook/notebook_model.js
@@ -211,14 +211,26 @@ Notebook.create_model = function()
             RCloud.UI.selection_bar.update(this.cells);
         },
         hide_selected_cells_results: function() {
-            _.each(this.get_selected_cells(), function(cell) {
-                cell.hide_cell_result();
-            });
+            if(this.get_selected_cells()) {
+              _.each(this.cells, function(cell) {
+                  cell.hide_cell_result();
+              });
+            } else {
+              _.each(this.get_selected_cells(), function(cell) {
+                  cell.hide_cell_result();
+              });
+            }
         },
         show_selected_cells_results: function() {
-            _.each(this.get_selected_cells(), function(cell) {
-                cell.show_cell_result();
-            });
+            if(this.get_selected_cells()) {
+              _.each(this.cells, function(cell) {
+                  cell.show_cell_result();
+              });
+            } else {
+              _.each(this.get_selected_cells(), function(cell) {
+                  cell.show_cell_result();
+              });
+            }
         },
         clear_all_selected_cells: function() {
             _.each(this.cells, function(cell) {
--- a/htdocs/js/notebook/notebook_model.js
+++ b/htdocs/js/notebook/notebook_model.js
@@ -210,6 +210,16 @@ Notebook.create_model = function()
             });
             RCloud.UI.selection_bar.update(this.cells);
         },
+        hide_selected_cells_results: function() {
+            _.each(this.get_selected_cells(), function(cell) {
+                cell.hide_cell_result();
+            });
+        },
+        show_selected_cells_results: function() {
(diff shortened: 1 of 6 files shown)


gordonwoodhull:
This works great and the implementation is clean (as usual!)

However, there are a couple of issues:
1. IMO there should be only one button, and it should toggle based on the result-state of the first selected cell. If that's too confusing (or hard to implement in our MVC), it should just hide and not show results. One can always show the results using the individual buttons on the cells. 
The problem with the current interface is that a disabled-looking button usually means the button is disabled, not that it disables something else. So I didn't think I could click on it. 
2. If there are no cells selected, the button should act on all cells.

useless5771:
Yes, I agree that the greyed-out icon is confusing, at one point I tried out the eye open and eye close icons https://www.w3schools.com/bootstrap/bootstrap_ref_comp_glyphs.asp but this was inconsistent with cells toggle-result icon. Another option was to use the eye icons for the cell hide/unhide toggle operation as well, but at the cell-level the eye icon can be confusing as it may suggest that whole cell will be hidden, not just the result...

How about having a single button which behavior changes when the control/command-key is pressed and by default it hides the selected cells? This way we can ensure that the feature is symmetrical - you can hide/show cells in bulk plus we just need a single icon.

Regarding what should happen if no cells are selected  - my only concern with this is that it may be confusing as the behavior of the hide/unhide button will be different to the other buttons on the selection-bar (delete, crop). The other buttons are controlled by the cells selection (and selection drop-down), but the hide/unhide will be controlled by cells selection, selection drop-down AND (when nothing is selected) it will act as if all cells were selected. If that is fine, I will implement this.

gordonwoodhull:
Having a single button with a modifier key sounds good. I don't think showing cells will be used all that much, as the main request was just to be able to see the code again when it gets lost in the output. (Note that running a cell always shows the output.) But you never know, once you put something out there people will find uses for it.

Commands like Export Notebook act on all cells if no cells are selected, and I think that should be the standard behavior. It's a special safety feature that delete and crop don't have any effect when nothing is selected, because those are dangerous commands. (Also it doesn't make much sense to crop to nothing, or delete nothing.) The selection drop down is just a shortcut and is equivalent to manually selecting cells.

Thanks!

useless5771:
Ok, cool, will apply the changes. 

Thanks

gordonwoodhull:
Oh, one other little change: I think it makes more sense to put the button either at the beginning or the end, since delete and crop are conceptually related to each other.

htdocs/js/notebook/notebook_model.js
```diff
@@ -211,14 +211,26 @@ Notebook.create_model = function()
             RCloud.UI.selection_bar.update(this.cells);
         },
         hide_selected_cells_results: function() {
-            _.each(this.get_selected_cells(), function(cell) {
-                cell.hide_cell_result();
-            });
+            if(this.get_selected_cells()) {
```
gordonwoodhull: I know it's annoying to deal with little nit-picky issues, but please test even the most trivial code.

`this.get_selected_cells()` always returns an array, which is truthy...

gordonwoodhull:
I'm moving the button after the select all/none menu but before the other buttons, and fixing two simple logic errors.

Thanks for your patience, @useless5771! UI work is an endless series of details... This works great apart from these slight issues.

PR #2413, opened 2017-05-10
Use readonly instead of disabled attribute on textarea

useless5771:
The 'disabled' attribute turns off the widget, the 'readonly' however makes it accessible so the text can be selected and copied.


gordonwoodhull:
Thanks @useless5771! Interesting distinction there.

Rebased and merged.

PR #2415, opened 2017-05-11
Fix #2361 by using actual cells languages when joining, not always R

### LINKED ISSUE #2361: joining python cell to markdown cell marks the code as R
It seems that any code is marked with `{r}` when joined to markdown, regardless of the source type.

useless5771:


gordonwoodhull:
This is one of those features that sounds really easy, but in practice it's fiendishly complicated. As I commented in the code when I foolishly implemented this the first time, it may even require some kind of multiple dispatch.

Your solution works for the case Markdown, R, Python, whether you join 1-2 then 2-3, or 2-3 then 1-2.

However it fails in different ways for R, Markdown, Python.

Since I can't reproduce backticks in Markdown, I'll use dashes below. If you join 1-2, 2-3, the first join works but the second produces 

```
---{r}
---{r}
3
---
1
---
---{python}
2
---
```

Maybe it has quoted the whole first joined cell 1-2 as R?

If you join 2-3 then 1-2, the text is correct:

```
---{r}
3
---
1
---{python}
2
---
```
but the language is set to R instead of Markdown.

I forgot what a colossal nuisance this feature is, when I assigned it to you. It is not a big deal to put it off till next version, since the bug has now existed for a few years.

useless5771:
Yes, my bad, silly oversight - I missed out the update of the cell language if markdown cell is merged into non-markdown cell.

useless5771:
I had another look at this and I added handling for situations when markdown and RMarkdown cells are joined.

So now the behavior is as follows:
* when two non-(r)markdown cells are joined, the result cell is markdown
* when markdown cell is joined with non-(r)markdown cell then the result is also markdown.
* when rmarkdown cell is joined with other cell, then the result is always rmarkdown

Now also when a cell is joined with (r)markdown cell that was already executed, the cell source gets displayed and result panel hidden, Before such cell would appear as empty.

gordonwoodhull:
Thanks @useless5771!

Congratulations! I think you nailed this one. At least I was unable to break it.

Am I right that this will probably work with arbitrary other languages, at least as far as RCloud language ids are valid Markdown language ids? Maybe it is not as many special cases as I feared, but just boils down to markdown vs non-markdown, and whether the languages are the same.

Rebased and merged for 1.9.1.

useless5771:
Yes, this should work with any language as long as they don't need to be handled in 'special' way as (r)markdown cells need to be.

An improvement worth considering would be to extract a Notebook Cell Joiner component, which would encapsulate this language-specific code and keep it out from notebook_controller. Thanks to this notebook_controller would be more generic.

PR #2416, opened 2017-05-16
Fix #2414 by not propagating clicks from dropdown

### LINKED ISSUE #2414: if you click in between recent notebooks, it minimizes the notebook tree
There's a gap between the items in the recent notebooks menu. That gap should probably go away, but what's worse, if you click on it (when no notebook is highlighted), it's apparently interpreted as a click on the panel header.

I was wondering why I kept on losing the notebook tree when trying to open a past notebook. I think this is it.

useless5771:
Also make links fill list items to remove space between elements in dropdown.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -1287,6 +1287,9 @@ var editor = function () {
                 e.preventDefault();
                 e.stopPropagation();
             });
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Great! Thanks @useless5771, rebased and merged

PR #2418, opened 2017-05-19
Fix #2417 by moving update_recent_notebooks to load_callback promise …

### LINKED ISSUE #2417: recent notebooks list is not updated for fork or create
Previously we refreshed the menu in the `editor_tab.load_callback`, which is called in every case where a notebook is loaded.

As of #2408, we refresh it in `load_notebook`, which only handles the case where a notebook was clicked in the tree or loaded from the query string. It doesn't update when a notebook is forked or created.

`update_recent_notebooks` should probably be more tightly bound to the `load_callback`, or it should be applied in every place where the `load_callback` is used.

useless5771:
…chain

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -1385,7 +1385,7 @@ var editor = function () {
                         RCloud.UI.fatal_dialog(message, "Continue", fail_url);
                         throw xep;
                     });
-                }).finally(that.update_recent_notebooks.bind(this));
+                });
         },
         open_notebook: function(gistname, version, source, selroot, new_window) {
             // really just load_notebook except possibly in a new window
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Okay, but this causes a regression of #2369, which is why we changed this in the first place. 😛 

Maybe there is no good place to put this.

useless5771:
Interesting, I tested this with a 'detached' notebook (notebook for which gist was deleted) and the Recent Notebooks list is populated. 

Do you have some other test case that I could follow to reproduce the regression?

gordonwoodhull:
Sorry, that ticket was really vague.

I'm referring to the case where you specify a bad notebook id in the URL, e.g. `?notebook=foo`

Then because you know that the Continue button causes a reload and is incredibly slow, you click Ignore and head for the recent notebooks. 😉 

But with this change the menu is completely empty.

I suppose an ugly solution is just to also populate them in the case where `load_notebook` fails.

gordonwoodhull:
I'll just take this myself, as it's pretty simple.

Great work on all the other issues, it's just `editor_tab` continues to be a confusing morass. Hopefully Shane's refactoring will help.

PR #2421, opened 2017-05-22
Add lib directory to considered paths, fixes att/rcloud.htmlwidgets#14

useless5771:


gordonwoodhull:
@useless5771, I don't understand why this still doesn't work for me. I apologize for not doing a deep investigation on this.

Given that 1.8 has already sailed, I am thinking that we should probably just put this off for a point release or 1.9.

The current URLs I'm seeing fail look like:

```
https://rcloud-5.research.att.com/shared.R/_htmlwidgets/gordon/parcoords/lib/d3/d3.min.js
https://rcloud-5.research.att.com/shared.R/_htmlwidgets/gordon/parcoords/parcoords.js
```

(There are 9 failing paths for this example.)

useless5771:
Yes this is odd, as the URLs now look correct, so there still must be something wrong in shared.R. Considering that the user-libs never actually worked, the code in shared.R that tries to handle them may not be correct.... I will have another look.

useless5771:
Ok, I was able to reproduce the problem by using the use.gist.user.home option.

Tested with notebook, view and flexdashboard endpoints.

As we discussed, notebooks using user-installed packages have limitations, e.g. they won't work when accessed anonymously or if the user accessing them don't have the dependencies installed in their own library.

gordonwoodhull:
I'm still not able to repro on my dev machine - even with `use.gist.user.home` and the package installed only in my personal libPath

URLs look like

```
http://127.0.0.1:8080/shared.R/_htmlwidgets/parcoords/lib/parallel-coordinates/d3.parcoords.js
```

and they.. work? I realize I could break them by installing https://github.com/att/rcloud.htmlwidgets/pull/17 but now I don't understand the underlying problem.

useless5771:
Hm... since these URLs work for you it looks like parcoords is still not installed in rcloud-managed user home directory - which should be somewhere here:

> {RCLOUD_ROOT}/data/home/{USER_NAME}/library/

Instead, I am guessing that parcoords is installed in local R libs directory located in your system home directory (i.e. /home/{username}/R/{version}/....) and this is why the URLs work (and you can't reproduce the issue).

If that is the case, try removing parcoords package from your local lib directory and installing it from rcloud notebook again (with use.gist.user.home setting turned on). Rcloud should then install parcoords to location like this:

> {RCLOUD_ROOT}/data/home/{USER_NAME}/library/parcoords/

(at least this is how it worked for me)

Yes, https://github.com/att/rcloud.htmlwidgets/pull/17 is required for user-installed packages to work in rcloud. Without it rcloud will keep producing URLs like:

> http://127.0.0.1:8080/shared.R/_htmlwidgets/parcoords/lib/parallel-coordinates/d3.parcoords.js

for resources in such packages, instead of URLs like this:

> http://127.0.0.1:8080/shared.R/_htmlwidgets/gordon/parcoords/lib/parallel-coordinates/d3.parcoords.js

(there is an extra path element holding a user name)

gordonwoodhull:
@s-u implemented `_htmlwidgets` with user/path much simpler way in c528db0d, but I think we may still need to add `lib`.

PR #2422, opened 2017-05-31
added selective import migration

ejen1:


s-u:
Thanks! Adds `--users <file>` option for selective migration and `--cmd <cmd>` for overriding the copy command (which defaults to `cp -pr`).

Issue #1645, opened 2015-08-19
file completion appends paths instead of replacing them


s-u:
![screen shot 2015-08-19 at 9 39 43 am](https://cloud.githubusercontent.com/assets/2073427/9357668/424914fa-4656-11e5-8172-b51c9f06ee8f.png)

Hit Enter:

![screen shot 2015-08-19 at 9 38 26 am](https://cloud.githubusercontent.com/assets/2073427/9357637/185b7bb0-4656-11e5-8911-072c924212d5.png)

s-u:
[BTW: note that the highlighting is wrong - the whole prefix should be highlighted instead of the random letters, and the superfluous slash]

gordonwoodhull:
I didn't know R would also match paths, and Ace is apparently unaware of this possiblity as well.

As noted in #1617, Ace is relying on a regex in order to search backward and match any replacement text. Right now I'm wondering if this should just be `/./` because it seems R is willing to replace just about anything.

s-u:
Well, R actually tells you exactly which part to replace (both the start point and the content) - so ideally, ACE shouldn't be guessing, but just replace what it's told to. It's impossible to use regexp to guess that since it depends on the context (symbols, files, ...) so the regexp would have to change.

So is there a way to tell ACE what to highlight and replace? Then we would simply use `$token` for the highlighting and `$start` to denote where to replace from.

gordonwoodhull:
yes, i am sure we could always dig deeper into the ace source........

s-u:
Heh, so what you mean is "no, there is no API for that" ...

gordonwoodhull:
Or, "I can't tell where the API ends and the library begins."

gordonwoodhull:
@shaneporter, I'm assigning you some wild ace hacking in a few issues. If this is too much or if I can offer pointers, please let me know. I've learned and forgotten this stuff a few times now.

shaneporter:
@gordonwoodhull I start typing a path in the prompt area, but the editor doesn't offer any suggestions. I guess I'm doing something wrong.

gordonwoodhull:
Completion is assigned to tab and ctrl-. - is this this missing clue?

shaneporter:
Great - got it. Thanks.

shaneporter:
The double slashes are coming back from R.

gordonwoodhull:
That's not the problem, though. The problem is that it's pasting a second copy of the whole path instead of completing it in place.

[...]

Issue #1617, opened 2015-08-14
code completion of a expression with brackets, copies earlier text


gordonwoodhull:
Similar to #646, when you ask for completion after a square bracket, R will show you your entire expression up until that point, even though the completion is really just on the subexpression following the bracket. 

Then because of the same regex-search-backward to find how much to substitute, which doesn't include bracket, the entire expression gets copied.

Not sure if there are other operators which have this IMO kind of strange behavior in R code completion, completing some larger expression than the context demands.

gordonwoodhull:
Example:
```{r}
a <- list(x=1, y=2, z=3)
```
![image](https://cloud.githubusercontent.com/assets/1366709/9282964/08bd133c-429e-11e5-88fa-38b2565b2790.png)

`a[a$` <tab>, substitutes immediately with `a[a[a$`, `a[a[a[a[a$` `a[a[a[a[a[a[a[a[a$`, ... `a[` * 2^n + `a`

gordonwoodhull:
Fixed in RCloud 1.9

Issue #645, opened 2014-05-07
auto-completion in the middle of something that exists, duplicates


gordonwoodhull:
This is really an ace bug, but the feature is pretty new and ace is nothing if not customizeable.

![image](https://cloud.githubusercontent.com/assets/1366709/2909812/e7a8c16a-d637-11e3-8ad7-5aff60e2814e.png)

![image](https://cloud.githubusercontent.com/assets/1366709/2909816/f59dcdd8-d637-11e3-98fb-56b3caea5a89.png)

It should recognize the text is already there and just point to after it.

gordonwoodhull:
Mats, since we're heading into the next version, why don't you take a look at this Ace editor code completion problem. It's related to #1635 and #1645. 

It's very important to fix these but they are a challenge because you have to wade through a lot of Ace code. You also have to understand the changes we've already made to Ace, which I've documented here:
https://github.com/att/rcloud/blob/develop/htdocs/lib/js/ace/ace.md

I recently had to re-apply all the changes, so I squashed them into one commit:
https://github.com/att/rcloud/commit/09e26791e8643a36feacfc1f855337c77b6eb5d1

So far these are just little tweaks - for autocomplete, the change just allows the completion call to be asynchronous. There's an attempt to fix the regex that Ace uses to search backward for the current token. But I think we'll have to to propose an optional feature for Ace so that it accepts the character positions to replace, instead of using the regex. 

We should know the character positions from the R completion call, but we're using too low-level an API right now:
https://github.com/att/rcloud/blob/develop/rcloud.packages/rcloud.r/R/main.R#L28-L35
Note all the private package members we're using - we need some of these but not all of them.

Here's the R's autocompletion documentation:
https://stat.ethz.ch/R-manual/R-devel/library/utils/html/rcompgen.html

PR #2430, opened 2017-06-20
Avoid duplication of inserted autocompletion

useless5771:
There are couple of changes in here, first commit modifies the code so r_mode uses ACE's Completer 'API' and modifies ACE core so it is possible to specify identifier matching regexp without modifying ACE's core.  With this change also the '/' character is included in the regexp matching the identifiers so paths completions are correctly handled - this partially fixes the #1645.

The second commit is the actual fix for #645 (and #1617) built on top of the above changes. It makes sure that:
* autocompletion is skipped if it resulted in duplication
* completion options that happen to start ahead of current position in the editor are correctly handled (#1617)

### DIFF
--- a/htdocs/lib/js/ace/mode-r.js
+++ b/htdocs/lib/js/ace/mode-r.js
@@ -110,10 +110,6 @@ define("ace/mode/r", function(require, exports, module)
          }
          return false;
       };
-      
-      this.getCompletions = function(state, session, pos, prefix, callback) {
-         this.$completer.getCompletions(state, session, pos, prefix, callback);
-      };
   
       this.lineCommentStart = ["#"];
    }).call(Mode.prototype);
@@ -130,9 +126,10 @@ var RCompletions = function() {
 };
 
 (function() {
-    this.getCompletions = function(state, session, pos, prefix, callback) {
+    this.getCompletions = function(editor, session, pos, callback) {
           var that = this;
-          rcloud.get_completions('R', session.getValue(), session.getDocument().positionToIndex(pos))
+          var line = session.getLine(pos.row);
+          rcloud.get_completions('R', line, pos.column)
                                  .then(function(ret) { 
                                    ret.forEach(function(x) { x.completer = that; }); 
                                   return ret;
@@ -142,14 +139,6 @@ var RCompletions = function() {
                                   });
       };
       
-   this.extractCompletionPrefix = function(editor, line, column, callback) {
-        var that = this;
-        rcloud.get_completion_prefix('R', line, column)
-                                .then(function(ret) {
-                                    callback(null, ret.value);
-                                  });
-   };
-   
    this.insertMatch = function(editor, completion) {
       var completions = editor.completer.completions;
       var session = editor.getSession();
--- a/htdocs/lib/js/ace/mode-r.js
+++ b/htdocs/lib/js/ace/mode-r.js
@@ -142,19 +142,26 @@ var RCompletions = function() {
                                   });
       };
       
-   this.identifierRegexps = [/[a-zA-Z_0-9:\/\$\-\.\u00A2-\uFFFF]/];
-      
-   this.insertMatch = function(editor, data) {
+   this.extractCompletionPrefix = function(editor, line, column, callback) {
+        var that = this;
+        rcloud.get_completion_prefix('R', line, column)
+                                .then(function(ret) {
+                                    callback(null, ret.value);
+                                  });
+   };
+   
+   this.insertMatch = function(editor, completion) {
       var completions = editor.completer.completions;
       var session = editor.getSession();
       var pos = editor.getCursorPosition();
+      var startPosition = completion.position;
       var line = session.getLine(pos.row);
-      var left = line.substr(0,pos.column);
+      var left = line.substr(startPosition, pos.column);
       var right = line.substr(pos.column, line.length);
       
-      var removeToLeft = function(editor, howMany) {
+      var removeToLeft = function(editor) {
           var range = editor.selection.getRange();
-          range.start.column -= howMany;
+          range.start.column = startPosition;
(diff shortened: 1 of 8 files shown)


gordonwoodhull:
This works! And I agree it's probably the best way to make Ace autocompletions work, within the constraints of how they designed the feature. And the refactoring helps.

But... the logic is so much more complicated than using the positions that R's autocomplete supplies.  

I'm worried that I don't know how to test this thoroughly, and also worried about the cost of maintaining this code.

Is the main objective to avoid designing a new API to deal with completers that supply the position?

useless5771:
Yes, I agree, the completion logic is more complex now, especially the code that performs the look-back. This is why I performed the initial refactoring and moved all R-specific logic to a single file so it is easier to upgrade ACE in the future and maintain the code.

We are limited here by ACE auto-completion API, we could return more information from R (for example), like:
* token that utils:::.guessTokenFromLine has 'guessed' and where it starts
* list of available completions for the guessed token

But... this would affect autocompletion functions in ACE (mode-r.js) anyways (as more information would be considered by the code that inserts the selected option). Also the problem with positioning of the autocomplete popup and completions happening ahead of current caret position in the editor would still need solving. Actually, the code using the 'extra' information may not necessarily be less complex - as the code would have access to: prefix (that ACE is using), filter (also coming from ACE), current caret position and extra information attached to autosuggestion options produced by R...

Another option I guess would be to refactor ACE (introduce an alternative completer API) that handles not only listing of suggestions but also decides on what is the position where the autocompletion popup should appear... This however will have a bigger impact on ACE core (and still the duplicate avoidance and second-step filtering that ACE applies to suggestions would need to be handled). This will also make it more difficult to upgrade ACE in the future...

I would suggest that we go with this change, it fixes couple of known issues with the autocompletion, it cleans up the way R support is registered in ACE and it is open for change in the future. Unless of course the solution in here doesn't work :) 

I guess I could spend some time investigating the latter option to at least identify what the impact of the refactoring would be, but I would definitely base the refactoring on top of commits in this PR.

gordonwoodhull:
I am afraid that this is too tricky and that there will be subtle bugs. It did work in the testing I did, so congrats for that!

htdocs/lib/js/ace/mode-r.js
```diff
@@ -116,7 +110,92 @@ define("ace/mode/r", function(require, exports, module)
...
+          var prefix = left.substr(lookbackIndex, left.length);
+          while(lookbackIndex < left.length && !data.value.startsWith(prefix)) {
+            lookbackIndex += 1;
+            prefix = left.substr(left.length-lookbackIndex, left.length);
+          }
+          if(left.length > lookbackIndex) {
+            removeToLeft(editor, left.length - lookbackIndex);
+          }
+        }
+      }
```
gordonwoodhull: This is the code which seems excessively complicated to me. 

Would you need to do any of this if you had the positions from R's autocomplete? It seems to me that if you just replaced the region defined by those positions, with the selected text, and put the cursor at the end, then there would be no need to search for anything or guess what was "meant" by a substitution.

gordonwoodhull:
@useless5771, could you please take a look into the impact of adding a more-informed completion API to Ace? I agree with most of the changes you've made but the cases to do with avoiding duplications and completion arriving to the right of the cursor seem unnecessary if the positions are known.

useless5771:
@gordonwoodhull with latest commit(s) the prefix and start position is provided by R completion provider and that information is then used to insert option selected by the user. 

The suggestions listed in the Autocomplete popup now come from single completer at a time, this is because Autocomplete popup only supports single prefix and prefixes are now identified by the completer itself and these can differ between enabled completers.

To switch between completers user just needs to hit 'CTRL+SPACE' when autocomplete popup is displayed and if there is some other completer that can provide suggestions, then suggestions from that completer will be listed. This behavior is similar to various IDEs (e.g. Eclipse)

The completers are queried in the same order as they are registered, which is: modeCompleter, keyWordCompleter, textCompleter, snippetCompleter.

where:
* modeCompleter queries the mode for completion options (in case of R, Python) it will query rcloud
* keyWordCompleter uses predefined list of keywords (this one is used by shell cells)
* textCompleter provides suggestions based on current cell content.
* snippetCompleter provides suggestions from SnippetManager

R completer now also provides completion options for function arguments and list elements (when '$' syntax is used). This is the bonus of merging the prefix and suggestions retrieval functionalities and letting R to do all completion resolution. 

I also made sure that requesting completions in Rmarkdown cell will not provoke error in console any more - this was in fact invalid configuration.

I extracted completer component in python language support so it also uses Completer API. And for now it should work as before (so it will use the prefix identified by the default utility function capturing identifier prefixes). 

I haven't made rcloud Python module to return prefix nor start position only because for some reason Python integration doesn't work in my rcloud instance (but it works when I invoke python from RStudio), I am getting this error when running python cells/python completion:
`
Error: get_completions: Error in py.eval(paste("runner = notebook_runner.NotebookRunner(rcloud_python_lib_path='",  : 
  Python exception: <type 'exceptions.OSError'> [Errno 9] Bad file descriptor
R trace:
.Call(rpy_eval, code, globals, locals, !quiet)
py.eval(paste("runner = notebook_runner.NotebookRunner(rcloud_python_lib_path='", 
    path, "', extra_arguments=['--matplotlib=inline'], executable='python')", 
    sep = ""))
.start.python(thissession)
`
But we can migrate python language module to new API with separate change...

gordonwoodhull:
Again, sorry this has taken so long to review and merge. However, in this case I don't think too much harm was done. This works almost perfectly - there are a couple of quirks which seem to belong to R or our configuration, but this definitely solves the problems it was designed for.

Wow, this is some serious engineering. Congratulations on figuring out how Ace does all this!

gordonwoodhull:
Rebased and merged to develop for 1.9

PR #2431, opened 2017-06-20
Resize autocomplete popup so it doesn't truncate options

### LINKED ISSUE #1635: really long completions get truncated in menu
image

useless5771:
FIX #1635


gordonwoodhull:
Thanks @useless5771! I wonder if they didn't expect completions to be this long... or maybe R is unusual in returning the entire path of the completion, rather than just the last token.

Anyway. Merged for 1.8.1!

PR #2432, opened 2017-06-20
Make highlighting of filtered options in autocomplete behave more intutitively

useless5771:
**Note** this pull request depends on changes from https://github.com/att/rcloud/pull/2430

ACE performs filtering of autocomplete options based on user's input by performing sort of position-based character search. The information acquired during filtering phase (which characters 'matched' and their score) are then used to display the options in the UI and highlight parts of an option which were identified as 'matching' the filtering criteria.  

This leads to the following behavior though:
For an input: '"/tmp/' + [TAB or CTRL+SPACE] + 'sth'
User will see (e.g.): ('*' - text between these is highlighted)
`
*/tmp/s*ome-pa*th*
`
`
 */tmp/*thi*s*-is-some-o*th*er-path
`

Although this can be intuitive for natural language completions, it isn't for file-system paths (which might have random characters) and tokens from programming languages. (Plus it seems that ACE's 
 filtering algorithm produces false positives)

This change makes sure that ACE just highlights the base continuous sequence of characters that matched the filtering criteria. E.g: (for the input from the above example)
`
*/tmp/s*ome-path
`
`
 */tmp/*this-is-some-other-path
`

### DIFF
--- a/htdocs/lib/js/ace/ext-language_tools.js
+++ b/htdocs/lib/js/ace/ext-language_tools.js
@@ -1081,6 +1081,7 @@ var AcePopup = function(parentNode) {
         }
         return tokens;
     };
+    
     bgTokenizer.$updateOnChange = noop;
     bgTokenizer.start = noop;
 
@@ -1338,8 +1339,12 @@ var Autocomplete = function() {
     };
 
     this.openPopup = function(editor, prefix, keepPopupPosition) {
-        if (!this.popup)
+        if (!this.popup) {
             this.$init();
+            if (editor.session.$mode && editor.session.$mode.$completer && editor.session.$mode.$completer.rowTokenizer) {
+                this.popup.session.bgTokenizer.$tokenizeRow = editor.session.$mode.$completer.rowTokenizer(this.popup);
+            }
+        }
 
         this.popup.setData(this.completions.filtered);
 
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Yeah, it's weird how they have two-pass completions, fetching results and then filtering through them with the characters typed (or arrowed through!) 

It's kind of cool, but also complicated.

However, I wouldn't just fix the symptom here. It's still going to search for non-contiguous characters, so I guess it's good that it highlights what it's doing, because otherwise you'd have no idea what is going on.

I think @s-u misunderstood what is happening [in his comment](https://github.com/att/rcloud/issues/1645#issuecomment-132603155). Because what is happening is confusing behavior.

I'd either leave it as it is, or force it only to match a contiguous prefix.

useless5771:
Yes, I agree this just fixes the symptom . Here I've chosen a solution that has a smaller impact on ACE core. Replacing the filtering algorithm would have a bigger impact as it isn't something that is specific to 'language mode'. But I will have another look anyways...

Thanks

gordonwoodhull:
Yeah I understand. It would be hard to fix what Ace is doing. So let's not obscure it by hiding the search our is performing.

Again, I think Simon misunderstood what was going on here. In the screen shot he gave, the tokenizing yielded a different target for replacement from what R wanted to replace. So it was highlighting based on the wrong input.

I don't think we should "fix" the highlighting.

s-u:
Well, ACE shouldn't be searching in the first place - it should ask for completion and deliver it, not guess. I care much less about the highlighting - main issue is the replacement being wrong. I'm just saying what is the expected behavior - regardless of what ACE actually does.

PR #2435, opened 2017-06-29
Remove rcloud.shiny-specific code

useless5771:
This also removes the generated css and css.map as these won't get removed when sass compiler runs.


gordonwoodhull:
If possible, this needs a temporary redirect from `shiny.html` -> `shared.R/rcloud.shiny/shiny.html`

It needs to preserve any query string parameters, of course.

People are going to have links to their old applications, which we'll need to support for a version or two.

useless5771:
Ok, thanks. I will have a look at redirects. The http.R in rcloud.support seems to be the place where we can introduce support for this.

gordonwoodhull:
I think you can use ordinary html redirects, but I'm not sure if those forward query string parameters, in which case you'd need to use javascript.

Either way, a short shiny.html replacement should do it, no need to delve into the server side.

useless5771:
Yes, I planned to send proper HTTP 301 response exactly because the content of meta http-equiv="refresh" would need to include query parameters. 

If we are happy with simple javascript-based solution I will go with this.

gordonwoodhull:
I see. That's annoying. Yes, there's no need for elegance here, it's just a temporary redirect.

gordonwoodhull:
rebased and merged for 1.8.1

gordonwoodhull:
Thanks @useless5771!

PR #2450, opened 2017-09-19
Add match case and match word options to find functionality, Fix #2425

### LINKED ISSUE #2425: Search whole word / case sensitivity
Shouldn't be hard to add, right? 

(Right?)

useless5771:
Added drop-down menu to find toolbar (next to the text input) which lets the user specify match case and match word options. 

If 'match word' option is selected the 'dot' character is treated as regular word character.

Cheers

### DIFF
--- a/htdocs/edit.html
+++ b/htdocs/edit.html
@@ -69,14 +69,21 @@
       <i class="icon-search"></i>
       <div class="find">
         <span id="find-details">
-          <input type=text id="find-input" class="form-control-ext mousetrap" placeholder="Find"></input>
-          <span id="match-status"><span id="match-index"></span> of <span id="match-total"></span></span>
+          <input type="text" id="find-input" class="form-control-ext mousetrap" placeholder="Find"></input>
+          <span id="match-status"><span id="match-index"/> of <span id="match-total"/></span>
+        </span>
+        <span class="btn-group" id="find-options-menu">
+            <button class="btn btn-default dropdown-toggle" data-toggle="dropdown"><i class="icon-ellipsis-horizontal"></i></button>
+            <ul class="dropdown-menu">
+              <li><span class="checkbox"><label><input type="checkbox" id="match-case-opt"/>Match case</label></span></li>
+              <li><span class="checkbox"><label><input type="checkbox" id="match-word-opt"/>Match word</label></span></li>
+            </ul>
         </span>
         <button id="find-last" class="btn" title="Find Previous"><i class="icon-arrow-left"></i></button>
         <button id="find-next" class="btn btn-primary" title="Find Next"><i class="icon-arrow-right"></i></button>
       </div>
       <div class="replace" style="display:none">
-        <input type=text id="replace-input" class="form-control-ext mousetrap" placeholder="Replace"></input>
+        <input type="text" id="replace-input" class="form-control-ext mousetrap" placeholder="Replace"></input>
         <button id="replace" class="btn">Replace</button>
         <button id="replace-all" class="btn">Replace All</button>
       </div>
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Thanks @useless5771!

The code looks good and works great, as always. One glitch I noticed is that the whole word mode doesn't match words at the beginning of the line, or the beginning or end of the cell

image

Changing `nonWord` to just `"\\b"` fixes this, as explained in one of the later answers here:

https://stackoverflow.com/questions/2951915/javascript-reg-ex-to-match-whole-word-only-bound-only-by-whitespace

(Strange that there was so much disagreement here!)

I'd also like some indication that something in the menu is selected. I like the trick of hiding these checkboxes in a menu, but it means that the current options aren't visible when the menu is closed. So I think that the background of the button for the menu should turn, say, light orange when any of the options is selected.

Finally, not a big deal but I found the close behavior for the menu to be inconsistent when clicking between it and the search input field (a common thing to do). Sometimes it would close and sometimes it would not. I think it should always close when it loses focus. I guess this might be an idiosyncrasy of how bootstrap dropdowns work, so I understand if you can't do anything about this.

useless5771:
Thanks @gordonwoodhull , I will have a look. Regarding the word boundary I must have missed that one, I haven't used \\b as it would also match the dot, but of course must have missed out other word boundary characters. I will update the regular expression.

Regarding inconsistent drop-down close behavior, yes, I noticed some other issues as well, e.g. when you click on next/previous buttons it wouldn't close as well. A lot of this is due to how some of the event handlers for the search form are implemented and prevent events from bubbling up and all look suspicious... Anyway, before spending time on investigating further the specific issues with dropdown behavior I wanted to know your feedback about proposed solution :)

Cheers

gordonwoodhull:
Ah yes I forgot about the focus handlers we already have there. Indeed they may not be the happiest but they are necessary - they are what remove the highlights when clicking away from the find dialog. 

If you can figure out a way to preserve that behavior while closing the drop-down, please go for it. I remember figuring out how to remove the highlights was a real headache for both Shane and me (and there may still be issues with Firefox) so if there's no good way then don't worry about it.

useless5771:
Added a commit which fixes dropdowns behavior when user clicks on buttons/text inputs in the find toolbar. Now not only find settings menu will be closed, but also other dropdowns in the UI (e.g. Recent Notebooks, Advanced etc...) which also suffered from the same issue.

I also made the find menu dropdown button to change its background color if any of the options is in effect. The color is 'orange'. 

Finally I have updated the regular expression matching whole words to simply use '\\b', tried to come up with a regular expression that would NOT match '.' as word boundary when a user requests whole words, but without using negative look-ahead and look-behind (which JS doesn't support apparently) I ended up with 100 character long regular expression which still didn't work...

gordonwoodhull:
Argh! Regular expressions are crazy sometimes.

I'm not sure if it's worth fixing this edge case, but JS does have lookahead with `x(?=y)`. Not sure if that's what you meant by negative look-ahead.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

useless5771:
Yes, lookahead (and negated flavor) is supported by JS, but lookbehind isn't.

so we can't use something like this: '\b(?<![\.])(text)(?![\.])\b'

gordonwoodhull:
Got it. Well, better to have some invalid matches than to end up with a regexp which is unmaintainable. I think R users can live with whole work matching parts of identifiers.

Thanks for your heroic effort!

gordonwoodhull:
Thanks @useless5771! 

The `stopPropagation` plus `close_bootstrap_dropdowns` is unfortunate, but I don't think it's worth wasting effort trying to understand and fix the deactivation of the find dialog, unless it becomes a problem.

Rebased and merged for 1.9.

PR #2455, opened 2017-10-17
Use DT to render variables in dataframe viewer

### LINKED ISSUE #2427: paginate the dataframe viewer
Users are reporting poor performance and outright crashing when using the dataframe viewer.

We should go ahead and paginate the thing.

Currently we chop it at 1000 rows, both on the server and client side. (Why - does the server side chop not work?) Instead we'd keep a reference to the whole thing on the R side, and just feed a user-specified number of rows, like 20, 50, 100 to the client.

It's probably easier to implement this ourselves than use DT, but we should also look at that package.

useless5771:
With this change DT (JS datatables) will be used to display variable values in dataframe viewer.
If a variable has data.frame type a pageable datatable will be rendered which only will send requested page to client.
If a variable doesn't have data.frame type a pageable datatable will be rendered, but all data will be sent to client.
    FIX #2427 

To achieve as simple and clear data viewer as possible I turned off some of the features that are enabled by default in DT: page size select box, search text input and I also used simple page navigation instead of the default that lists first, last and neighboring page numbers. Alternative to turning these features off would be to modify CSS so the viewer is more compact.

BTW The server-side processing feature of DT used in this solution could also be exposed in flex.dashboard/RCAP/view modes allowing for displaying large data frames.

### DIFF
--- a/rcloud.packages/rcloud.viewer/inst/javascript/rcloud.viewer.js
+++ b/rcloud.packages/rcloud.viewer/inst/javascript/rcloud.viewer.js
@@ -28,33 +28,8 @@ return {
         k();
     },
     view: function(data, title, k) {
-        $('#viewer-body > table').remove();
-        var columns = data.r_attributes.names;
-        if(typeof columns === 'string')
-            columns = [columns];
-        columns = _.without(columns, 'r_attributes', 'r_type');
-        if(columns.length<1)
-            k();
-        var nrows = data[columns[0]].length;
-        if(nrows>1000) nrows = 1000;
-        // styling the table would go better with CSS but we can only
-        // install CSS by URL right now(?)
-        var header_style = 'border: 0; background-color: #dedede; font-family: sans-serif; font-size: 12px; text-align: right; white-space: nowrap';
-        var datum_style = 'border-style: solid; border-width: thin 0; border-color: #aaa; text-align: right; padding-left: 10px; white-space: nowrap';
-        var header = $.el.tr({}, [$.el.th({style: 'border: none'})].concat(columns.map(function(x) {
-            return $.el.th({scope:'col', style: header_style}, x);
-        })));
-        var rows = [header];
-        for(var i = 0; i<nrows; ++i) {
-            function fetch(col) {
-                return $.el.td({style: datum_style}, data[col][i]);
-            }
-            var items = [$.el.th({scope: 'row', style: header_style}, i)].concat(columns.map(fetch));;
-            rows.push($.el.tr({}, items));
-        }
-        $('#viewer-body').append($.el.table({style: "border-collapse: collapse; max-width: none"}, rows));
-        $('#viewer-body tr:first-child + tr td').css({'border-top': 0});
-        $('#viewer-body tr:first-child th').css({'padding-left': '10px'});
+        $('#viewer-body > div').remove();
+        $('#viewer-body').append($(data));
         RCloud.UI.right_panel.collapse($("#collapse-data-viewer"), false, false);
         k();
     }
--- a/rcloud.packages/rcloud.enviewer/R/rcloud.enviewer.R
+++ b/rcloud.packages/rcloud.enviewer/R/rcloud.enviewer.R
@@ -2,12 +2,42 @@ rcloud.enviewer.refresh <- function()
   rcloud.enviewer.on.change(.GlobalEnv)
 
 ## OCAP
-rcloud.enviewer.view.dataframe <- function(expr)
-  View(get(expr, .GlobalEnv))
+rcloud.enviewer.view.dataframe <- function(expr) {
+  # pass un-evaluated expression, so it can be used by View
+  View(get(expr, .GlobalEnv), expr = expr)
+}
+
+#'
+#' @param name data.frame variable name
+#' @param options dataTables request parameters
+#' @return datatables JS result object (see datatables documentation)
+rcloud.enviewer.view.dataframe.page <- function(name, options) {
+  val <- get(name, .GlobalEnv)
+  
+  result <- list(draw = options$draw)
+  
+  if(is.data.frame(val)) {
+    records <- nrow(val)
+    result$recordsTotal <- records
+    result$recordsFiltered <- records 
+    page <- seq(options$start,min(options$start+options$length, records))
+    data <- tryCatch(jsonlite::toJSON(cbind(as.integer(rownames(val)[page]),val[page,]), dataframe = "values"))
+    if(typeof(data) == "try-error") {
+      result$error <- paste0("Could not marshal data.frame data. Error: ", as.character(data))
+    } else {
+      result$data <- data
+    }
+    return(result)
+  } else {
+    result$error <- paste0("Unsupported variable type: ", typeof(val))
+  }
+  return(result)
+}
 
 ## -- how to handle each group --
-rcloud.enviewer.display.dataframe <- function(x, val)
+rcloud.enviewer.display.dataframe <- function(x, val) {
   structure(list(command="view", object=x, text=paste0("data.frame [",paste(dim(val), collapse=', '),"]")), class="data")
+}
 
 rcloud.enviewer.display.value <- function(val) {
     type <- class(val)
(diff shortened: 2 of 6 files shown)


rcloud.packages/rcloud.viewer/inst/javascript/rcloud.viewer.js
```diff
@@ -28,33 +28,8 @@ return {
...
-                return $.el.td({style: datum_style}, data[col][i]);
-            }
-            var items = [$.el.th({scope: 'row', style: header_style}, i)].concat(columns.map(fetch));;
-            rows.push($.el.tr({}, items));
-        }
-        $('#viewer-body').append($.el.table({style: "border-collapse: collapse; max-width: none"}, rows));
-        $('#viewer-body tr:first-child + tr td').css({'border-top': 0});
-        $('#viewer-body tr:first-child th').css({'padding-left': '10px'});
+        $('#viewer-body > div').remove();
+        $('#viewer-body').append($(data));
```
gordonwoodhull: This line is throwing for me. We must have different jQuerys or something, because for me `$(data)` returns a string and then jQuery/Sizzle crashes trying to look at the `ownerDocument` of that string, which of course has none. This happens whether I do `View(iris)` or `i <- iris` and then click on `i` in the Workspace.

gordonwoodhull: DT 0.2 from CRAN. `data` is 

```js
[
  "list(filter = \"none\", caption = \"<caption>Data: i</caption>\", data = list(` ` = character(0), Sepal.Length = logical(0), Sepal.Width = logical(0), Petal.Length = logical(0), Petal.Width = logical(0), Species = logical(0)), container = \"<table class=\\\"display\\\">\\n  <thead>\\n    <tr>\\n      <th> </th>\\n      <th>Sepal.Length</th>\\n      <th>Sepal.Width</th>\\n      <th>Petal.Length</th>\\n      <th>Petal.Width</th>\\n      <th>Species</th>\\n    </tr>\\n  </thead>\\n</table>\", options = list(paging = TRUE, \n    pagingType = \"full\", lengthChange = FALSE, searching = FALSE, info = FALSE, ordering = FALSE, pageLength = 20, serverSide = TRUE, ajax = \"function(data, callback, settings) {\\n    if(window.parent && window.parent.RCloud && window.parent.RCloud.UI && window.parent.RCloud.UI.enviewer) {\\n        window.parent.RCloud.UI.enviewer.dataFrameCallback(\\\"i\\\", data, callback, settings);\\n    }\\n    }\", order = list(), autoWidth = FALSE, orderClasses = FALSE, columnDefs = list(list(orderable = FALSE, targets = 0))))",
  "100%",
  "NULL",
  "list(defaultWidth = NULL, defaultHeight = NULL, padding = NULL, viewer = list(defaultWidth = NULL, defaultHeight = NULL, padding = NULL, fill = TRUE, suppress = FALSE, paneHeight = NULL), browser = list(defaultWidth = NULL, defaultHeight = NULL, padding = NULL, fill = FALSE), knitr = list(defaultWidth = \"100%\", defaultHeight = \"auto\", figure = FALSE))",
  "list(list(name = \"dt-core\", version = \"1.10.12\", src = list(file = \"/usr/local/lib/R/site-library/DT/htmlwidgets/lib/datatables\"), meta = NULL, script = \"js/jquery.dataTables.min.js\", stylesheet = c(\"css/jquery.dataTables.min.css\", \"css/jquery.dataTables.extra.css\"), head = NULL, attachment = NULL, package = NULL, all_files = FALSE))",
  "NULL",
  "function (instance) \n{\n    data = instance[[\"x\"]][[\"data\"]]\n    if (object.size(data) > 1500000 && getOption(\"DT.warn.size\", TRUE)) \n        warning(\"It seems your data is too big for client-side DataTables. You may \", \"consider server-side processing: http://rstudio.github.io/DT/server.html\")\n    data = escapeData(data, escape, colnames)\n    data = unname(data)\n    instance$x$data = data\n    instance\n}",
  "list()"
]
```

useless5771: Hm.. interesting, I will have a look and try to reproduce. Maybe indeed I use some different versions of libraries.

useless5771: Couldn't reproduce so far, 'iris' dataset got displayed without any problems using View(iris) and from Workspace view. This is the output of sessionInfo in my environment.

`
R version 3.2.3 (2015-12-10)
Platform: x86_64-pc-linux-gnu (64-bit)

locale:
[1] LC_CTYPE=en_US.UTF-8 LC_NUMERIC=C
[3] LC_TIME=en_US.UTF-8 LC_COLLATE=en_US.UTF-8
[5] LC_MONETARY=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8
[7] LC_PAPER=en_US.UTF-8 LC_NAME=C
[9] LC_ADDRESS=C LC_TELEPHONE=C
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C

attached base packages:
[1] parallel stats graphics grDevices utils datasets methods
[8] base

other attached packages:
[1] DT_0.2 rcloud.shiny_0.5
[3] rcloud.rcap_0.4.5.2 rcloud.sh_0.1-1
[5] rcloud.python_0.0.1 rcloud.flexdashboard_1.0.1
[7] rcloud.rmd_1.0.0 rcloud.htmlwidgets_1.0.3
[9] rcloud.notebook.info_0.1 rcloud.enviewer_0.1
[11] rcloud.viewer_0.1 rmarkdown_1.4
[13] githubgist_1.0-3 rediscc_0.1-4
[15] sendmailR_1.2-1 RCurl_1.95-4.8
[17] bitops_1.0-6 httr_1.2.1
[19] markdown_0.7.11 knitr_1.15.1
[21] png_0.1-8 Rserve_1.8-5
[23] FastRWeb_1.1-2 Cairo_1.5-9
[25] rcloud.support_1.8-0 rjson_0.2.15
[27] base64enc_0.1-4

loaded via a namespace (and not attached):
[1] Rcpp_0.12.10 rcloud.r_0.0.1 tools_3.2.3
[4] unixtools_0.1-1 digest_0.6.12 uuid_0.1-2
[7] gist_1.0-0 jsonlite_1.5 evaluate_0.10
[10] PKI_0.1-5 yaml_2.1.14 curl_2.5
[13] stringr_1.2.0 htmlwidgets_0.8 flexdashboard_0.5
[16] rprojroot_1.2 R6_2.2.2 Rook_1.1-1
[19] rcloud.web_1.0-3 github_0.9.8 magrittr_1.5
[22] codetools_0.2-14 backports_1.0.5 htmltools_0.3.6
[25] mime_0.5 brew_1.0-6 stringi_1.1.5
[28] rcloud.rmarkdown_0.0.1
`

I will rebase the commits onto current develop branch and make sure that I have the latest htmlwidgets and other packages installed... And that DT is from CRAN as well..

useless5771:
Just tried this out on develop and hotfix/1.8.1 branches with reinstalled DT from CRAN and both work fine in my environment. Also tried on FF and Chrome.

BTW, this is the value of 'data' that I get, at that point (Actual DT table HTML page removed as I couldn't get it to format nicely:

`<div class="rcloud-htmlwidget-content" id="rc_htmlwidget_content_541181"><iframe frameBorder="0" width="100%" height="400" srcdoc="<!DOCTYPE html>[[[[ HTMLWIDGET HTML CONTENT ]]]]"></iframe></div>`

useless5771:
got it! Just removed rcloud.htmlwidgets package and I reproduced the error that you get:

```javascript
jquery-2.1.1.js:1409 Uncaught TypeError: Cannot read property 'ownerDocument' of undefined
    at Function.Sizzle.contains (jquery-2.1.1.js:1409)
    at Function.buildFragment (jquery-2.1.1.js:5123)
    at jQuery.fn.init.domManip (jquery-2.1.1.js:5363)
    at jQuery.fn.init.append (jquery-2.1.1.js:5194)
    at view (rcloud.viewer.js:32)
    at WebSocket.socket.onmessage (rserve.js:1040)
```

The dataframe viewer requires rcloud.htmlwidgets to work , I will add a check which will prevent that error from being thrown and display appropriate error message in the viewer instead.

useless5771:
I added rcloud.htmlwidgets package to suggested packages by rcloud.viewer package and made the View function to handle situation when rcloud.htmlwidget package is not installed by displaying an error message.

What is odd though is that you don't have rcloud.htmlwidgets package installed, unless this is not the case, then I will investigate this issue further.

gordonwoodhull:
Ah! Fantastic, thanks for tracking that down and adding a reasonable error message.

Yes, this is a new development machine, so I still need to install the basics.

gordonwoodhull:
Looks great! Let's run with this.

Rebased and merged for RCloud 1.9

PR #2456, opened 2017-10-18
Introduce scrolling to recent notebooks dropdown menu, FIX #2423

### LINKED ISSUE #2423: Scrollable recent notebooks
Recent notebooks is an extremely useful feature, but it is currently limited to a fixed number of notebooks. It would be nice if it could scroll to show more notebooks when desired.

useless5771:
Initially the recent notebooks dropdown is populated with twenty items and if there are more notebooks available in the history an extra menu item option is added to the bottom of the menu, which when clicked loads next batch of twenty notebooks.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -1661,7 +1661,8 @@ var editor = function () {
                 .then(this.populate_recent_notebooks_list.bind(this));
         },
         populate_recent_notebooks_list: function(data) {
-            var recentNotebooksBatchSize = 20;
+            var firstRecentNotebooksBatchSize = 10;
+            var recentNotebooksBatchSize = 2*firstRecentNotebooksBatchSize;
             var transformData = function(data) {
               return _.chain(data)
                 .pairs()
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
For some reason, the menu is considerably shorter than the number of items, at least on the machine I'm currently testing (Mac). There are 12 items showing but if you know it's there, it scrolls to display 8 more plus the arrow. 

image

It definitely should show all the way up to and include the arrow, so that users know this feature is available.

gordonwoodhull:
Otherwise, it works great!

useless5771:
Looks like difference in browsers behavior, I tested on Chrome and it displays vertical scrollbar from the start. I will have a look.

Thanks

gordonwoodhull:
This is Chrome/MacOS. 

Macs never display scrollbars until you start scrolling. So that explains why it was not apparent I could scroll. This causes a lot of annoyance when doing cross platform UI. I think if the arrow is showing then it will be apparent enough.

I notice you're using absolute pixel sizes. It might actually be tricky to get it to show up to and including the arrow on all platforms. Maybe set the page size shorter and let the browser calculate the menu size? I'm not sure.

useless5771:
Ok, so I have reduced the size of first batch to ten items, this way the arrow is displayed first time the dropdown is displayed. I also added a tooltip to the arrow so it is clear what clicking on it will do.

I tried an alternative approach which would make the arrow always visible and only the list of recent notebooks would be scrollable, but ensuring that bootstrap menu styling is preserved required copying a lot of CSS, so not great from maintainability point of view...

Regarding pixel sizes, we need to define the maximum height at which the scrollbars should be displayed. I updated the max-height property to appropriate value in em units.

gordonwoodhull:
This works great. Thanks @useless5771!

Issue #2448, opened 2017-09-13
tree filtering based on time


gordonwoodhull:
@s-u mentioned the idea of seeing only the past week's notebooks, past month's, etc.

I think we should do #2387 chronological sort of the tree first, but after that let's look at applying time-based filtering of the tree.

shaneporter:
I assume that the filtering options should be derived from the `last_commit`s of available tree items, and update when new items are loaded (or saved).

(and also show the number of items that will be shown?)

e.g:

Previous 7 days (107 of 208)

shaneporter:
I'm assuming that the current notebook should always be shown in the tree, regardless of whether it matches the date filter.

However, what happens if a user opens a notebook that doesn't match the filter, and then opens another. Should that previous notebook then be hidden?

I'm just wondering whether hiding the nodes is the best approach, and that some other approach isn't preferable.

gordonwoodhull:
I am assuming we should keep the current behavior of just continuing to open folders, and leave closing of folders to the user. Suppose the user is looking for all notebooks on a certain topic, for example, and then they intend to do some comparison.

shaneporter:
Following a discussion with you @gordonwoodhull, I'll look to implement a 'if it's in, it's shown'. (Future filter changes will update the tree, but the current notebook is *always* shown).

gordonwoodhull:
(deleted my confused comment above)

Yes - certainly the current notebook should always be shown. 

We're undecided whether this means that things should be shown that don't match the filter, or the filter should be switched off if the user opens a notebook which doesn't match. Go with whichever is easiest and we'll iterate.

gordonwoodhull:
This is complete - we're just working out the bugs now.

Issue #2391, opened 2017-03-10
incremental search for notebook titles


gordonwoodhull:
The notebook tree is growing out of control and incremental search is a fast way to prune it, at least when you know the title. 

(If you don't know the title, we could do SOLR-based pruning of the tree, but that's slower and not appropriate for incremental search. Likewise the tree could also be filtered using notebook recommendations. I mention these features to motivate the design; they won't be part of this release.)

IMO the ideal way to add incremental search is without introducing another mode.

It should apply to both

1. the current notebook tree interface
2. the chronologically sorted notebook tree(s) of #2387

I'm open to discussion about the right interface. I am sure I am missing some subtleties here.

My first thought is for search not to affect which trees are open. Matching is against the path as well as notebook name; thus if you search for a folder name and it shows up, you can click the folder to open it, and all of its notebooks are shown because they meet the filter.

But it may be that we want to open trees in order to show matches. I'm not sure about this.

Search can be started with a shortcut (e.g. <kbd>ctrl s</kbd>) or by clicking a button in the notebook pane header. It inserts a small text-entry control above the tree. The filtering will stop when either the search string is empty, or the incremental search is closed with <kbd>esc</kbd> or clicking a close-x.

s-u:
+1 cool idea

shaneporter:
I think that the incremental search should not modify the structure of the tree (by hiding items), nor highlight or scroll to the current matching folder or notebook.

It should focus entirely on the matches, showing information about whether the match is based on the notebook name or the path.  

The user should be able to scroll through each match using the up/down cursor keys, and select a match with either enter or a mouse click. When a match is selected, the incremental search should be closed and the appropriate notebook or folder should be opened.

gordonwoodhull:
@shaneporter, I guess you are proposing some sort of overlay that's separate from the tree? 

That's also valid, but what's wrong with hiding parts of the tree? Is this to make the implementation simpler?

It's been a while but I think the reason I thought it should happen in the same tree is partly so that it could combine with the chronologically sorted "tree". Also it is still a tree, at least in non-chronological mode, so it might not be that much simpler to display it separately.

shaneporter:
Yes, that's exactly it. My inspiration came from my current editor of choice, Visual Studio Code, which like Sublime Text and no doubt lots of other editors has an incremental search much like this:

![image](https://user-images.githubusercontent.com/2493614/30280935-ec496e32-9708-11e7-94f5-53ef1706881a.png)

[...]

Issue #2387, opened 2017-02-26
time-based notebook tree


gordonwoodhull:
Imagine keeping the top level folders (Notebooks I Starred, People I Starred), but at the author level,  just show all of their notebooks in reverse chronological order. 

Because folders stop making sense, we probably need to show each notebook with its full path, abbreviated somehow.

gordonwoodhull:
Nay, folders still make sense. Just the same tree we have now, but folders and notebooks are sorted by  modification date - or last-run date once we have the data. The date of a folder is most recent date of any of its contents.

shaneporter:
I'm going to go with the folder date as being the most recent date of any of its _immediate_ children, like OS X finder.

gordonwoodhull:
@shaneporter, sorry I missed this comment earlier. 

I waited until merging everything before testing this out, so that I could see it in context. I find it somewhat confusing.

My intuition is that the most recent notebooks should float to the top, no matter how deep they are in the folders. In other words, that latest date should be defined recursively.

I see what you mean about OSX but I think that behavior might be by accident - it's not that they propagate the date to the immediate folder but that many programs actually modify the folder (by creating and deleting files) when saving a file.

gordonwoodhull:
I'll open another issue for discussion. This feature is complete and it's useful either way.

PR #2457, opened 2017-10-18
Review for 'tree refactor': filtering, ordering and incremental search

shaneporter:
#2448 - tree filtering based on time
#2391 - incremental search
#2387 - time sorted notebook tree

I apologise that this is such a huge PR (and that many of the earlier commits are described simply as 'refactor'); I am aware of some issues, but also suspect that some other areas may not be bulletproof:

- Incremental search's SOLR query needs tweaking, but hopefully what's returned _is_ displayed correctly, and users can open the notebooks
- Imported notebooks may not display if they don't match the current filter


htdocs/js/tree/notebook_tree_model.js
```diff
@@ -0,0 +1,1691 @@
...
+
+            // initial assignment: 
+            this.tree_data_ = data;
+
+            this.update_filter({
+                prop: 'tree_filter_date',
+                value: opts['tree_filter_date']
+            });
+
+            this.update_sort_type(opts['tree_sort_order'], true);
```
gordonwoodhull: remember that these `opts` will be undefined the first time the code is run - probably worth testing with all options set to `null`...

gordonwoodhull:
As can be expected, I got pulled into some other "urgent" work and wasn't able to review this week.

I think I will just merge once it works well enough, since this is too huge to wait much longer. I trust that you know what you're doing, and it will just take time to absorb all these changes.

Once I got around the error above by manipulating data in the debugger, sort and filter worked great in a quick test! Excellent work, @shaneporter.

The layout could use tightening up. This may be a Chrome/MacOS issue but there is some line wrapping which makes it tall.

image

I'd expect it also to stick to the top and not scroll, but I'm open to opinions on that. I think if it is just one line tall as intended, it won't take up too much space. When digging in All Notebooks it's quite common to be scrolled away, and we wouldn't want to forget what the sort/filter are.

I'm not sure if the white border helps - guess this will be clear when it's the right height.

It seems strange to me to pop up a dialog for incremental search? Shouldn't that just be another line inside the notebook tree? And of course I'm missing some search APIs so I can't test it.

Merge next week is going to be a lot of fun!

gordonwoodhull:
Ah, I guess the dialog is because we are not incrementally searching within the tree UI. Hmm. I'm not a huge fan of modal dialogs, but I'm not sure what to do. May be too late to do anything about this now.

Also, <kbd>cmd p</kbd> is the print hotkey, which we should not steal. Can we use <kbd>alt s</kbd>?

gordonwoodhull:
This works almost perfectly, and we're running out of time, so I'll go ahead and merge today.

I'll let you know separately about any issues.

I like the new design of the incremental search dialog, too - attaching it to the top makes sense.

Thanks @shaneporter!

gordonwoodhull:
@shaneporter, @dougmet, I'm slightly confused about merging this and #2459, since they both modified search.R but I know you guys were working together on this.

What I will try to do (and I know it's too late in the day/week to get any feedback until Monday, so I'm partly posting this so that I remember what I did):

1.  Keep all of Doug's changes, which remove search.R and put that stuff into rcloud.solr
2. Also keep `rcloud.search.description` (restoring search.R) but change it so that it references the rcloud.solr package instead of directly calling `.solr.get`

It looks like `rcloud.search.description` is a bit higher-level than the rest of rcloud.solr, so it should stay in rcloud.support, but any opinions are welcome here.

@s-u and I are in the same place today so I'm hoping to get our dev server up and running.

Thanks!

gordonwoodhull:
Ah, okay, I was just confused in the middle of the merge, everything fits together now!

shaneporter:
@dougmet  wrote the `rcloud.search.description` method for the incremental search so I had something to call, but I anticipated there being some changes from 'my' side once the merge had been done. Thanks.

gordonwoodhull:
Thanks, I think I understand. It is moved into rcloud.solr. 

I just wasn't finished with the merge when I thought I was!

PR #2472, opened 2017-11-09
Open notebook in view mode when alt-clicked from recent notebooks…

### LINKED ISSUE #2424: view mode for recent notebooks
It would be nice to be able to open a notebook form the recent notebooks list in view mode

useless5771:
FIX #2424

I had to do some tweaking of ubuntu desktop settings to test this one, as by default ubuntu is capturing alt-clicks (It triggers window drag and move functionality).

But this raises a question if we should consider changing the alteration key to something else than <kbd>alt click</kbd>, as this might be a common problem for ubuntu users?

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -450,7 +450,12 @@ var editor = function () {
                 e.preventDefault();
                 var gist = $(e.currentTarget).data('gist');
                 $('.dropdown-toggle.recent-btn').dropdown("toggle");
-                result.open_notebook(gist, undefined, undefined, undefined, e.metaKey || e.ctrlKey);
+                if(e.altKey) {
+                  var url = ui_utils.make_url('view.html', {notebook: gist});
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I'd be open to changing the key on Ubuntu. This works great for me on Cinnamon Mint but I think I also disabled <kbd>alt drag</kbd> for moving windows around.

The sticky thing is that <kbd>ctrl</kbd> and <kbd>shift</kbd> are taken; the browser uses those for open in new tab or new window. <kbd>ctrl shift click</kbd>?

BTW, we should be adding new keyboard shortcuts to the shortcuts help using `RCloud.UI.shortcut_manager` - I think I missed this on some of your earlier contributions and I'll have to revisit that at some point.

gordonwoodhull:
Drat, it looks like I accidentally merged this. I'll open a few issues to fill in the gaps.

PR #2480, opened 2017-11-10
Notebook tree nodes sort order select initialised on load

shaneporter:


gordonwoodhull:
Thanks @shaneporter, rebased and merged.

PR #2481, opened 2017-11-10
Notebook tree date period filter set and applied correctly

shaneporter:


gordonwoodhull:

PR #2486, opened 2017-11-10
Check if solr feature is enabled where needed. FIX #2479

### LINKED ISSUE #2479: do not throw an error if no rcloud.solr
Making a notebook visible or hidden gives me

```
Error: set_notebook_visibility: Error in loadNamespace(name) : there is no package called ‘rcloud.solr’
R trace:
stop(e)
value[[3L]](cond)
tryCatchOne(expr, names, parentenv, handlers[[1L]])
tryCatchList(expr, classes, parentenv, handlers)
```

because I haven't installed rcloud.solr. But I also don't have solr.url and it shouldn't be trying to update solr.

useless5771:
Backend not always was checking if solr feature is enabled. Introduced a function that not only checks if solr.url property is set but also if rcloud.solr package is available and added solr feature checks where they were missing.

### DIFF
--- a/rcloud.support/R/notebook.comments.R
+++ b/rcloud.support/R/notebook.comments.R
@@ -11,7 +11,7 @@ rcloud.post.comment <- function(id, content)
 {
   res <- create.gist.comment(id, content, ctx = .rcloud.get.gist.context())
   rcloud.comments.email(id, content, ' posted a new')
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
Thanks @useless5771! Rebased and merged.

Issue #2475, opened 2017-11-09
history doesn't show up


gordonwoodhull:
Apparently for notebooks with a lot of history, it will show the ellipsis, but otherwise it just opens up a few pixels and shows nothing.

![no history](https://user-images.githubusercontent.com/1366709/32630607-410e2f80-c56b-11e7-9138-c77641af6909.png)

PR #2487, opened 2017-11-10
Use configured view to open notebook from recent list

useless5771:
Added a new method resolve_view_link to RCloud.UI.share_button extension which can be used to resolve URLs of view pages for given notebooks.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -451,8 +451,9 @@ var editor = function () {
                 var gist = $(e.currentTarget).data('gist');
                 $('.dropdown-toggle.recent-btn').dropdown("toggle");
                 if(e.altKey) {
-                  var url = ui_utils.make_url('view.html', {notebook: gist});
-                  window.open(url, "_blank");
+                  RCloud.UI.share_button.resolve_view_link(gist).then(function(url) {
+                    window.open(url, "_blank");
+                  });
                 } else {
                   result.open_notebook(gist, undefined, undefined, undefined, e.metaKey || e.ctrlKey);
                 }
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Thanks @useless5771 - could you please refactor a bit so that this code is shared with `set_page`? 

Ideally `resolve_view_link` could take over the version/tag stuff that `set_page` is doing (because next thing you know someone will want to resolve a view link with a version/tag), and then `set_page` could call `resolve_view_link`.

It makes me antsy to have code duplicated like this. I know I'm often guilty of it when I'm in a hurry, but I think we are doing okay for this release so let's get it right.

useless5771:
Thanks, rearranged the code so it is clear what it does and removed code duplication.

gordonwoodhull:
For future reference, the paths to do with version can be tested by going to a historical version of the notebook.

gordonwoodhull:
Thanks @useless5771! The code is much more clear this way. 

The path where there was a version was incomplete, I fixed that in a631a7537f505963. 

Rebased and merged.

useless5771:
Forgot to state that I couldn't test this part, sorry about that. In my dev instance the history of notebooks is most of the time just '...' (if it appears) and clicking on '...' does nothing. I thought this was just temporary regression related to solr and/or filtering feature, but I just updated to the latest 'develop' branch state and I still just get '...' (or nothing).

It appears that there must be something wrong with my instance and I will need to investigate this. @gordonwoodhull , have you experienced similar behavior?

gordonwoodhull:
Ah right that is UI bug #2475. I was able to test only by using the undo key <kbd>ctrl alt z</kbd> to go back a version.

Ain't integration fun?

useless5771:
yep, indeed is fun :) thanks

PR #2489, opened 2017-11-10
view.html no longer crashes with get_notebook_info error

shaneporter:
This was all too apparent after resolving the initial 'view' issue (relating to incremental search).


gordonwoodhull:

PR #2492, opened 2017-11-14
History nodes show up

shaneporter:
When I made the change to fix the history nodes, I found some other issues, such as history nodes reordering and disappearing when the date filter was updated. This PR resolves those other issues as well as the original fault.


gordonwoodhull:
Yay history is back!

Good catch, those are some weird feature interactions. Yes, I can't think of any reason we'd want history nodes to be affected by either filtering or sorting.

PR #2493, opened 2017-11-14
Changing sort order should be faster

shaneporter:
When I originally thought of the sort order changes, I mistakenly thought that perhaps passing in a list of parent/child nodes would be faster, but the DOM updates were a huge bottleneck (as you found @gordonwoodhull). I overthought this, and now pass in the entire tree, and from my tests, things seem to be faster. I hope that you see a significant improvement on your 'notebook heavy' tree.

### DIFF
--- a/htdocs/js/tree/notebook_tree_model.js
+++ b/htdocs/js/tree/notebook_tree_model.js
@@ -902,7 +902,7 @@ notebook_tree_model.prototype = {
                 update_children(this.tree_data_);
 
                 this.on_update_sort_order.notify({
-                    nodes: nodes_and_children,
+                    tree_data: this.tree_data_,
                     sort_type: sort_type
                 });
             }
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Yes! This does it. I only have 209 notebooks in my tree but the time dropped from 5.2s to 1s.

Will update when I push to the dev server where there are many more.

gordonwoodhull:
Okay, I find these performance metrics hard to account for. On the research lake, where I have slightly fewer notebooks (203) but there are hundreds of users, the time drops from over 2 minutes (with the fan starting up at 30 seconds) to 2 seconds. 

Why would the number of users affect this, when none of their notebook trees are open?

My vague memory is that every update to a jqTree actually refreshes the whole thing. There are no incremental changes.

Anyway, can't argue with results!

gordonwoodhull:

PR #2494, opened 2017-11-14
Empty folders hidden on date filter

shaneporter:
If the folder contains no matching descendant notebooks it will be hidden.

### DIFF
--- a/htdocs/js/tree/notebook_tree_model.js
+++ b/htdocs/js/tree/notebook_tree_model.js
@@ -34,6 +34,7 @@ var notebook_tree_model = function(username, show_terse_dates) {
 
     this.sorted_by_ = this.orderType.DEFAULT;
     this.matches_filter_ = [];
+    this.empty_folders_ = [];
 
     // functions that filter the tree:
     this.tree_filters_ = {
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Hi @shaneporter - I accidentally merged this at the end of the day Tuesday, without testing it thoroughly. 

As you noticed, it broke a lot of things. For example:
- renaming a notebook and forking throw errors
- notebook history is gone again

I've reverted the entire PR in 46acfe51402. Please have a look at what went wrong.

gordonwoodhull:
One last thing big thing is broken by this PR, confusing enough that it makes it hard to test.

The notebook trees `/People I Starred/My Notebooks` and `/All Notebooks/My Notebooks` are not populated on page load - only `/Notebooks I Starred/My Notebooks`.

It looks like all notebooks under that level, that is, with user-defined folders, are populated correctly, it's just the user-level notebooks that are missing.

After changing the date filter to Last 7 or 30 Days and back to All, the notebooks are shown.

If reloading the page when a non-starred notebook is current, this can make it look like all notebooks are missing at the user's top level.

gordonwoodhull:
Thanks @shaneporter!

BTW, I think you had said there were issues with notebook names not updating between the various trees, but I wasn't able to repro. So maybe I misheard you.

shaneporter:
That’s what I said, but I can no longer reproduce either!

PR #2505, opened 2017-11-18
Scroll to node after import/recent selection

shaneporter:
The fixed header for filtering, sorting caused the issue where the `offsetTop()` call for the first argument to `scroll_into_view` would always return 0.

Updated the selector to the element that scrolls and the correct value is returned and scroll to node works as expected.


gordonwoodhull:
Yay! It's kind of scary how much I depend on this one.

Rebased and merge. Thanks @shaneporter!

PR #2506, opened 2017-11-20
Open forked notebook in new tab on ctrl-click, FIX #2441

### LINKED ISSUE #2441: cmd-click on "fork" kills the current session
`<Cmd>`-click on "fork" is expected to open a new tab with the forked notebook, but instead it kills the current session and replaces it with a new session and the forked notebook.

useless5771:


gordonwoodhull:
Yay! Now I just wonder how many other places we should be doing this.

I was surprised, but not all that surprised, how big a change this was. You are right that we are very far from the single responsibility principle. If you have any suggestions for how to refactor this particular tangle of fork/star/update/load, they are welcome!

Rebased and merged for 1.9.

Issue #2501, opened 2017-11-18
new folder is put at the bottom in order-by-date mode


gordonwoodhull:
When the notebook ordering is by date and renaming a notebook so that it creates a folder, the folder is placed at the end instead of the beginning.

This also happens if the notebook tree is unknown, e.g.:

1. Open an old notebook that is the only one in its folder, where the folder is somewhere near the middle or top when sorted by date.
2. Copy its notebook id.
3. Delete the notebook. You'll be sent to the last opened notebook.
4. In the URL, paste the notebook id over the current one and press enter.
5. When page finishes loading, all the notebooks and folders will be in order, but that folder will be at the end.

gordonwoodhull:
In the second case, the folder should be sorted into the tree based on the date of the notebook.

Issue #2497, opened 2017-11-14
define date of folder recursively


gordonwoodhull:
My intuition is that when notebooks are sorted by date (#2387), the date of folders should be the date of the most recent notebook, determined recursively.

Currently the date of a folder is the date of the most recent immediate notebook, but I think that the most recent notebook should appear at the top of the tree, no matter how deeply nested it is.

Folders don't really exist in RCloud - in a file system there is another object which has its own date, determined by the last time anything was added or removed from the folder. But folders in RCloud are only groupings of notebooks with the same prefix.

I'm open to arguments here - have to admit I hadn't really thought about it until I saw it.

gordonwoodhull:
I think the implementation would be

1. When reading the tree initially, calculate and save the latest date of all folders from the bottom up. 
2. Whenever a change is made to a notebook date, traverse from the notebook up the tree and take the max date of the saved value vs the change date .

Usually a change to a notebook is because the user changed it, so it will be the max date, but the metadata can get out of sync with the backend, so it is also possible for a notebook to be updated to some old date.

gordonwoodhull:
Also please consider the order of "new old" folders that pop up when a notebook is new to this instance but has a date which places its folder in the middle of the sort, e.g. https://github.com/att/rcloud/pull/2507#issuecomment-346152914

PR #2507, opened 2017-11-20
New folder appears in correct date sort location

shaneporter:
My code didn't account for newly created folders, which have neither `children` nor a `last_commit` property. The resulting date was incorrect, and resulted in the last place ordering by time.

Referring to the issue, #2501, this change means that newly created folders (resulting from a notebook rename) are put at the top of the date ordering. 

The second case, where a notebook in a folder is deleted and loaded by ID, results in the folder being at the top of the tree.

### DIFF
--- a/htdocs/js/tree/notebook_tree_model.js
+++ b/htdocs/js/tree/notebook_tree_model.js
@@ -303,9 +303,16 @@ notebook_tree_model.prototype = {
                 return lc;
 
             } else {
-                var dateA = a.last_commit ? new Date(a.last_commit) : _.max(_.map(a.children, function(child) { return new Date(child.last_commit); }));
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
But it shouldn't always go at the top:
image

This folder should be sorted in based on its notebook date. A rare case perhaps, unless importing/visiting a bunch of notebooks which don't exist in this RCloud instance.

gordonwoodhull:
Ping @shaneporter, if you can order the new folder by the date of its notebook, that would be super.

Otherwise, if you'd rather handle this in 1.9.1 as part of #2497, that's also acceptible. Just let me know so I can merge this and go ahead with the release.

shaneporter:
If it's ok with you, I'll handle it as part of #2497.

gordonwoodhull:
Sure, that's fine. Thanks!

gordonwoodhull:
Good for now. Thanks @shaneporter! Rebased and merged.

PR #2509, opened 2017-11-20
Pull and Replace close clears fields

shaneporter:
Dialog closes, fields are cleared. More reliably this time.


gordonwoodhull:
But now the pull button doesn't close the dialog, which is alarming because the user sees the field cleared but it's actually remembered.

shaneporter:
In fe407e6, I renamed `close_dialog` to `reset_dialog` and close the dialog explicitly after a successful pull and replace.

gordonwoodhull:
That'll do it! Thanks @shaneporter!

Rebased and merged.

PR #2512, opened 2017-11-21
Hidden notebook icon colour

shaneporter:


gordonwoodhull:
Kind of idiotic to use the same class name for the button as for the state, isn't it? Who could have done that?

Oh, me of course. I'm an idiot. a4772eb4dab5a6

gordonwoodhull:
Instead of complicating things further, I renamed the button classes to `hide-notebook` and `show-notebook` instead of `shown-notebook` and `hidden-notebook`, which made no sense whatsoever.

gordonwoodhull:
Thanks for figuring this out, @shaneporter!

PR #2514, opened 2017-11-23
Fix resizing of data frame viewer panel, FIX #2503

### LINKED ISSUE #2503: dataframe doesn't share space with assets pane correctly
There is a rather arcane algorithm in `RCloud.UI.collapsible_column.resize` that is supposed to share the space when the requests from the panels on either side exceed the window height.

It isn't getting invoked or isn't being supplied the correct data for the dataframe viewer, so if both the assets pane and dataframe viewer are showing on an average laptop screen, the dataframe viewer will go off the bottom.

It's not hard to workaround by hiding the assets, but the user shouldn't have to do that. Compare what happens when the asset pane and help pane are both open.

useless5771:
This change makes sure that:
* height of htmlwidget iframe is provided to the resizing algorithm, instead of null.
* htmlwidget iframe hard-coded height does not override whatever value panel resizing algorithm sets on the panel (this was resulting in iframe overflowing the panel)

### DIFF
--- a/rcloud.packages/rcloud.enviewer/R/rcloud.enviewer.R
+++ b/rcloud.packages/rcloud.enviewer/R/rcloud.enviewer.R
@@ -7,33 +7,6 @@ rcloud.enviewer.view.dataframe <- function(expr) {
   View(get(expr, .GlobalEnv), expr = expr)
 }
 
-#'
-#' @param name data.frame variable name
-#' @param options dataTables request parameters
-#' @return datatables JS result object (see datatables documentation)
-rcloud.enviewer.view.dataframe.page <- function(name, options) {
-  val <- get(name, .GlobalEnv)
-  
-  result <- list(draw = options$draw)
-  
-  if(is.data.frame(val)) {
-    records <- nrow(val)
-    result$recordsTotal <- records
-    result$recordsFiltered <- records 
-    page <- seq(options$start,min(options$start+options$length, records))
-    data <- tryCatch(jsonlite::toJSON(cbind(as.integer(rownames(val)[page]),val[page,]), dataframe = "values"))
-    if(typeof(data) == "try-error") {
-      result$error <- paste0("Could not marshal data.frame data. Error: ", as.character(data))
-    } else {
-      result$data <- data
-    }
-    return(result)
-  } else {
-    result$error <- paste0("Unsupported variable type: ", typeof(val))
-  }
-  return(result)
-}
-
 ## -- how to handle each group --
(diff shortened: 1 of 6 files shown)


gordonwoodhull:
The sizing works great. However, sometimes the size of the `#DataTables_Table_0_wrapper` div ends up too short, making it look like the dataframe is empty. (As a guess, maybe the height is calculated based on the empty dataframe?)

There is no way to recover once this happens - even if new data is displayed, it is cropped out. 

datatable too small

I haven't been able to narrow down exactly when it happens - I thought it was when the Dataframe was initially hidden, but currently I am unable to get it to work by having it shown, so that's not it.

useless5771:
I've seen this happening at one point when I worked on dataframe viewer, but this was when iframe content was not refreshed after showing the panel and/or there was something wrong with container hierarchy. 

I tried reproducing this, with no luck, tried various scenarios:
* resizing the right-hand side panel (with dataframe panel expanded and collapsed)
* resizing browser window
* different sets of panels collapsed/expanded in the right-hand side panel, with content being displayed in the dataframe viewer and without
* various ways of displaying dataframe viewer (via workspace panel or using the 'View' function)
* various sizes of the data frames
* the same data.frame displayed in the editor cell using DT and in dataframe viewer
* collapsing all panels and bringing them back

Next step would be to provoke latency in various places to see if there are no race conditions or such.

Did that happen more than once?

The workaround would be to re-execute 'View(iris)' or clicking the data.frame from workspace panel.

gordonwoodhull:
It happens consistently for me now. It doesn't matter how many times I run `View`, directly or indirectly - once it is in this state, it simply will not display data.

I'll have to test on other machines. So far I've just been testing locally on Linux with Chrome & Firefox

gordonwoodhull:
A little bit more detail, based on what we discovered in our working session: if the Workspace panel (rcloud.enviewer) is open at the first time `View` is called, directly or indirectly, then the data frame shows correctly. Otherwise the communication between the front-end and back-end seems to be broken, and no data is displayed. This is consistent across Linux and MacOS for me.

gordonwoodhull:
FWIW, I doubt this is related, but `size_this` in rcloud.htmlwidgets (htmlwidgets.js) is spinning and using significant CPU, whether or not the dataframe successfully shows. (Just 1-6%, but enough to get the fan running on my ThinkPad.) 

It appears that `Di.is('visible')` is always false.

gordonwoodhull:
Yep, that was it. Thanks @useless5771! Rebased and merged for 1.9

PR #2525, opened 2017-12-20
Fixes namespace issues for event/tree/search/date filter

shaneporter:
So that third-party JS that RCloud packages may rely upon are not affected by window scoped objects.


gordonwoodhull:
Thanks @shaneporter! Rebased and merged for 1.9.1

PR #2526, opened 2018-01-08
Autoscroll output in notebook pane when cell is executed.

### LINKED ISSUE #2222: autoscroll output in the notebook pane
We should do the same as in the session pane and autoscroll the notebook if the bottom line of the running cell is visible.

useless5771:
The scroll will only be performed if the executed cell's input is visible and output would not fit into displayed area in the notebook pane.

FIX #2222 

I also investigated what would be involved in applying different autoscroll strategies depending on:
* if 'Run All' was selected
* if the prompt cell was visible when bulk of cells got executed.
and in such case autoscroll to cell outputs despite visibility of their source.

Supporting the above would require a bit more work, as the view would need to be aware of the state of the view when execution got triggered and use it when results are rendered to apply autoscrolling. 

Implementing these however do have some implications on user experience - e.g. when a user runs all cells, the notebook pane would scroll until all cells got processed, this may not necessarily be user-friendly if there are a few htmlwidgets in the notebook. Also we could potentially want to disable autoscrolling during execution of all cells, if user manually scrolls the notebook pane. 

@gordonwoodhull, what do you think? Should I have a look at this?

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -774,6 +774,7 @@ function create_cell_html_view(language, cell_model) {
             }
             this.state_changed(error ? 'error' : running_state_==='unknown-running' ? 'ready' : 'complete');
             current_result_ = current_error_ = null;
+            this.scroll_to_result();
         },
         clear_result: clear_result,
         set_readonly: function(readonly) {
@@ -927,6 +928,19 @@ function create_cell_html_view(language, cell_model) {
                 source_div_.show();
                 edit_button_border(true);
             }
+        },
+        should_scroll: function() {
+            return ui_utils.is_visible_in_scrollable($('#rcloud-cellarea'), [notebook_cell_div, source_div_]) &&
+                   !ui_utils.is_visible_in_scrollable($('#rcloud-cellarea'), [notebook_cell_div, result_div_]);
+        },
+        scroll_to_result: function() {
+            var that = this;
+            ui_utils.on_next_tick(function() {
+                var cellarea = $('#rcloud-cellarea');
+                if(result_div_ && that.should_scroll()) {
+                  ui_utils.scroll_to_after(result_div_, undefined, cellarea, [notebook_cell_div], cellarea.height());
+                }
+            });
         },
         toggle_results: function(val) {
             if(val===undefined)
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -30,6 +30,7 @@ function create_cell_html_view(language, cell_model) {
     var highlights_;
     var code_preprocessors_ = []; // will be an extension point, someday
     var running_state_;  // running state
+    var running_ui_state_ = {add_result: {}};
 
     // input1
     var prompt_text_;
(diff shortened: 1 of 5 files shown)


gordonwoodhull:
Let's discuss tomorrow on the call. I wasn't aware that there would be so many considerations. But I guess I'm not surprised, since the notebook interface is already so complicated.

In particular, I thought that the case where the user has manually scrolled would already be handled by design, because the notebook pane should only autoscroll if the final lines are visible.

gordonwoodhull:
Also I'm not clear what it would have to do with the input of the cell - it's mostly about the end of the output being visible but in danger of disappearing.

gordonwoodhull:
No question, this is a tricky feature. 

One use case for this (probably the main use case) is long-running cells that are doing a lot of work. For that reason, you'll need to hook in deeper than `end_output` - probably `add_result` would be more appropriate.

Consider the following test case:

```r
for(i in 1:10) {
    print(1:100)
    Sys.sleep(2);
} 
```

After each output, the notebook should scroll.

I still have to chew on the rest of your comments.

useless5771:
Thank you for providing example use case, it clarified the requirements a lot.

BTW, the example that you provided results in RCloud session crash in my env. I am guessing it works fine for you? 

I will workaround this for now by using this:
```{r}
res <- lapply(c(1:10), function(x) {
    print(1:100)
    Sys.sleep(2)
} ) 
```
 which does pretty much the same thing.

gordonwoodhull:
Hmm, that is probably the odd rserve.conf settings that you figured out to get pandoc working? Wild! I have no idea why a for loop would be different than lapply.

useless5771:
Tried with both configurations - default and 'tweaked' - and it failed... odd...  

New commit holds improved autoscrolling implementation, which supports the example use case and the simple use cases. 

I guess the question is if we should add support for the use cases covering execution of multiple cells in this thread, or should we raise a separate ticket? This might simplify change management...

gordonwoodhull:
@useless5771, are you perchance trying to use RCloud with R3.4? 

So far we are only compatible with 3.3; segfaults will occur with 3.4.

gordonwoodhull:
This works great now.

I am beginning to understand the issues you raised above.

However think there might be simpler solutions.

 1. It's not exactly about whether Run All was selected, but about whether there are more cells enqueued to run. So perhaps what we could do is just scroll to each cell as it is started, as long as the end of results of the last cell are visible.

 2. Or alternately, maybe we could just scroll to the **next** cell, so that if you run cells in sequence, the next one can always "capture" the scroll.

However, I agree this is another level of complication and we should take small steps.

Could you please put this under an option e.g. 'Autoscroll notebook', so that if it drives people crazy they can turn it off? This is the behavior that I usually want but there are other use cases where it might be annoying, particularly if you want to enqueue multiple cells.

useless5771:
Ok, I will add the option to disable the autoscroll. 

Regarding R version, indeed I use  3.4.3 (I upgraded it when I was investigating the issues with python/rmarkdown cells). I will downgrade it. Thanks.

useless5771:
"Autoscroll notebook" option exposed.

useless5771:
Downgrading R also fixed the for loop issue, thanks.

gordonwoodhull:
Thank @useless5771! This works great. I was even able to disable and reenable the feature while a cell was running, manually scroll down to the new end of output, and autoscrolling picked right up again.

Since the multicell stuff is tricky and it's unclear what we want to do there, let's take small steps and see how people react to the one-cell version. Hopefully we can find a simple and elegant way to implement the multicell version of the feature based on the experience we gain from this.

Rebased and merged for 1.9.1.

PR #2527, opened 2018-01-09
Introduce 'stop' button

### LINKED ISSUE #2442: Clicking on Stop can result in Play All killing the session
Clicking on "Stop" can result in "Play All" in case where the browser lags due to massive output. The problem with that is that when the user wanted to desperately save the session by stopping the output, the entire session is blown away instead by the inadvertent "Play All" - a really bad user experience.

It's unclear to me how to prevent this other than not using the same button for Stop and Play All.

useless5771:
Split functionality that was implemented by run_button into three components:
* run_button - handling run button actions
* stop_button - handling requests to stop processing
* processing_queue - holding currently processed cells queue.

The refactoring also removes a cyclic dependency between run_button and cell_controller (through shell_tab, notebook_controller intermediaries).

Also avoid double-scheduling of notebook cells due to unintentional double-clicks and/or race conditions between scheduling process and user clicking the Run button multiple times.

FIX #2442


gordonwoodhull:
This works great, @useless5771! Excellent idea to split up the functionality into three classes.

One thing I noticed in a quick test is that, while Stop is properly disabled when things stop running, it should also be disabled when the page is initially loaded.

Also, we noticed that the existing highlight is unclear, I opened #2530.

gordonwoodhull:
Thanks @useless5771! Rebased and merged for 1.9.1.

PR #2528, opened 2018-01-09
Add colour coding to recent notebooks dropdown

### LINKED ISSUE #2462: color coding of recent notebooks
Now that we have a sort of infinite scrolling for recent notebooks, it would be helpful to see at a glance how old the notebooks are.

It would be really clever to color-code the background based on how old the notebook is, e.g. green is new, yellow recent, orange, brown. This would be especially helpful if it detected gaps in the dates, so that all notebooks from three days ago are one color and from five days ago are another color, i.e. not a linear mapping of colors but something that detects "runs".

A challenge/question here is whether we need to reassign colors as more items are shown in the menu. That might be disorienting, but on the other hand, we'll run out of colors eventually if we don't.

useless5771:
Introduced five styles that control background of items in recent notebooks dropdown, these styles are then used by a styling function when recent notebooks dropdown is populated.

I used lighter colours than the ones suggested in the ticket, as simple 'orange', 'yellow' and others seemed too 'invasive', as these are controlled by CSS we can easily change them:

image

FIX #2462

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -5,7 +5,8 @@ var editor = function () {
         github_nonfork_warning = ["GitHub returns the same notebook if you fork a notebook more than once, so you are seeing your old fork of this notebook.",
                                   "If you want to fork the latest version, open your fork in GitHub (through the Advanced menu) and delete it. Then fork the notebook again."].join(' '),
         NOTEBOOK_LOAD_FAILS = 5,
-        tree_controller_;
+        tree_controller_,
+        color_recent_notebooks_by_modification_date_;
 
     function has_notebook_info(gistname) {
         return tree_controller_.has_notebook_info(gistname);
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
As discussed on the call, let's add the option to color code by either last-opened date (the default) or last-modified date (what's currently implemented). As such, the setting in the settings pane could be called "Color recent notebooks by modification date" (if that fits). 

Lmk if you run into problems creating the setting (or if you find the process for creating settings could be optimized). It should just be a matter of adding an entry in `htdocs/js/ui/settings_frame.js`

useless5771:
Thanks for pointing me to the setttings_frame.js.

The last commit adds color coding based on the last access time. The implementation assigns different colors based on the time difference between current date and last access time, scaled by the time difference between current date and oldest notebook included in the dropdown.

This of course reassigns colors depending on how many notebooks are displayed in the dropdown, what may be confusing at first. This solution however captures user 'sessions', e.g. notebooks viewed on given day, or couple hours ago, will have the same background. Requesting more notebooks then may be interpreted  as change of the 'perspective', if you'd like.

Alternative I guess could be to introduce mapping based on hardcoded rules, e.g.:
* notebooks from today - always green
* notebooks 1d-3d old - light green
* notebooks 3d+2 weeks old - yellow
* notebooks 2 weeks + 1 month old - orange
* older notebooks - purple

In such case however, if someone opens up RCloud after longer break, they may end up with all notebooks in the same color - e.g. purple...

gordonwoodhull:
This is fantastic, especially along with the taller menu from #2537. I'm just adding the flag that displays "Reload to see changes" when this option is changed.

Thanks @useless5771! Rebased and merged for 1.9.1

PR #2531, opened 2018-01-10
pass params.username for use in rcloud.solr search description

### LINKED ISSUE #2517: incremental search by username does not seem to have any effect
As soon as I start typing in the username field, I get a lot of notebooks, but they don't seem to match the username pattern.

If the notebook name pattern is filled in, then typing in the username field has no effect on filtering.

bashlee:
Username missed from call. 
Update also made to rcloud.solr to incorporate username 
Fixes #2517


gordonwoodhull:
Great! Thanks @bashlee! Rebased and merged for 1.9.1.

PR #2535, opened 2018-01-11
Highlight button using icon color.

### LINKED ISSUE #2530: make play button highlight "more clear"
Now that we have a Stop button, the cloudy "highlight" on the play button seems "unclear" to me. 

I think this is because it's active but it's dimmer than the stop button which got activated.

Let's animate it to (I think) a pulsing yellow and remove the fuzzy highlight thing. jQuery has animation functionality which we use elsewhere. Should be a one-liner to change the highlight.

useless5771:
It took much longer than I thought. I am also not happy with how it works/looks as the animation is affected by updates to the page while a notebook is executed. I wonder if we should consider using progress spinner instead of animating the color. This would be consistent with the way the progress is indicated on notebook cells when they are running. @gordonwoodhull what do you think?

Regarding implementation, it simply animates the color of the 'play' icon.

I tried using jquery-ui animations so the highlighting style would be specified in CSS, but for some reason animations submitted by jquery-ui are not executed. Are you aware of that? Are they disabled on purpose or is this a bug?

FIX #2530

### DIFF
--- a/htdocs/js/ui/navbar.js
+++ b/htdocs/js/ui/navbar.js
@@ -42,13 +42,32 @@ RCloud.UI.navbar = (function() {
         },
         create_highlight_button: function(id, title, icon) {
             var result = this.create_button(id, title, icon);
-            result.control = $.el.span($.el.span({
-                class: 'button-highlight'
-            }), result.control);
+            result.highlight_color = "#ffff4d";
+            result.icon_class = icon;
+            result.original_color = null;
+            result.highlighted = false;
             result.highlight = function(whether) {
-                $(result.control)
-                    .find('.button-highlight')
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
I think it looks pretty darn good. Sorry if it was frustrating to implement - that's pretty common for fiddly UI stuff!

I found that the yellow wasn't prominent enough, so I added a delay step.

If once I knew about jquery-ui animations, I have forgotten. 

I see what you mean about it being affected by changes to the page, but I don't think we can do much about this. Even the progress GIF stutters when the page is doing a lot of work. IIUC the renderer is essentially single-threaded.

gordonwoodhull:
This is rebased and merged for 1.9.1

PR #2536, opened 2018-01-11
Update handling of cells with unknown states

### LINKED ISSUE #2529: editing a running cell causes its output to disappear
Create a slow cell (for example with a `Sys.sleep` in it). Run it, and then start editing the code.

Once autosave kicks in (and status turns to spinning question mark), the output is irretrievably lost and it isn't clear what's going on.

Strangely, it doesn't seem like output is lost if a non-running cell is run, so I'm not sure what's happening here. (It could be intentional, but if so it's a bad idea.)

useless5771:
Make sure that result div of running cells is not cleaned up on edit so the output is displayed.
Also use 'unknown' state as the final state for such cells, not 'Ready'.

FIX #2529

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -678,8 +678,10 @@ function create_cell_html_view(language, cell_model) {
         },
         state_changed: function(state) {
             var control = left_controls_.controls['run_state'];
-            if(running_state_==="unknown" && state==="running")
+            if(running_state_==="unknown" && state==="running") {
                 state = "unknown-running";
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I probably just spent about as much time reviewing this as you spent writing it. This logic is very confusing. 

I think these fixes are 100% correct. There are probably other cases that are wrong - for example, why does changing a running cell put it in the unknown cell, while changing an already-run cell puts it in the ready state?

Ah well, this fixes the lost output and that's the most important thing!

Rebased and merged for 1.9.1. Thanks @useless5771!

PR #2537, opened 2018-01-12
Make recent notebooks dropdown height relative to screen size

### LINKED ISSUE #2532: recent notebooks menu should be closer to the height of the window
Showing only 10 recent items is constricting when the screen size is a lot larger. 

The "page size" should be dynamically determined based on the height of the window. The idea is to be able to quickly select a notebook that is on the tip of your tongue, so the more shown the better.

I think the right length is the greater of 2/3 the window height, or 50px from the bottom of the window.

This requires determining the height of an item. I think the easiest method would be to include one dummy item in the source html. Then select that item and use `element.getBoundingClientRect().height` to get its height.

useless5771:
FIX #2532

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -418,8 +418,46 @@ var editor = function () {
         },
         populate_recent_notebooks_list: function(data) {
             var that = this;
-            var firstRecentNotebooksBatchSize = 10;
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Very nice. Thanks @useless5771! Rebased and merged for 1.9.1.

PR #2539, opened 2018-01-18
Use has_notebook_info instead of checking if property is set.

### LINKED ISSUE #2538: new notebooks are hidden?
I don't understand how I never saw this before, or how no one has reported this. 

Do people like it this way?

In RCloud 1.8 a new notebook was shown (visible to others). In RCloud 1.9 it's hidden and you have to click the eye to make it shown.

useless5771:
FIX #2538

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -143,7 +143,7 @@ var editor = function () {
                     before = Promise.reject(new Error(
                         "Invalid gist source '" + source + "'; available sources are: " +
                             gist_sources.join(', ')));
-                } else if(!get_notebook_info(gistname).username) {
+                } else if(!has_notebook_info(gistname)) {
                     set_notebook_info(gistname, { source: source });
                     before = rcloud.set_notebook_property(gistname, "source", source);
                 }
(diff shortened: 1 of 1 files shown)


htdocs/editor_tab.js
```diff
@@ -195,8 +195,11 @@ var editor = function () {
...
                 .then(shell.new_notebook.bind(shell))
                 .then(function(notebook) {
-                    tree_controller_.set_visibility(notebook.id, true);
-                    that.star_notebook(true, {notebook: notebook, make_current: true, version: null});
+                    return tree_controller_.set_visibility(notebook.id, true).then(function(x) {
+                        return notebook;
+                      });
+                }).then(function(notebook) {
+                  return that.star_notebook(true, {notebook: notebook, make_current: true, version: null});
                 });
```
gordonwoodhull: Is this part of the fix? These were intentionally in parallel before, I will probably change to `Promise.all()` before merging.

gordonwoodhull:
Aha, I see it was continued fallout from the notebook tree MVC work - I hadn't fixed the existence check correctly.

Thanks @useless5771! Will merge later today.

PR #2540, opened 2018-01-19
Attach source file to parsed cells

useless5771:
Shiny 'showcase' mode uses srcfile information to trace back the code that is executed so appropriate parts of code can be highlighted. Without this information the source file reference is always set to '<text>' and code lookup in Shiny dashboard fails

This change makes sure that the parsed cell code has the source file attached.


gordonwoodhull:
Excellent, thanks @useless5771!

Looks like this is potentially helpful for any debugging functions, not just Shiny's showcase mode.

I tried out the same change in `rcloud.packages/rcloud.r` and it enabled source files in the native R debugger.

I'll open a separate PR for that, since there is an API change and I'm not sure of the consequences.

gordonwoodhull:
Rebased and merged for 1.9.1

Issue #2420, opened 2017-05-19
rcloud.delete.asset doesn't work


gordonwoodhull:
Somehow the JSON ends up encoded wrong. Running the last line of  `rcloud.upload.asset` in the debugger, but without `invisible`:

```R
Browse[2]> foo <- rcloud.update.notebook(notebook, list(files = l))
Browse[2]> foo
$ok
[1] FALSE

$content
$content$message
[1] "Invalid request.\n\nFor 'properties/content', nil is not a string."

$content$documentation_url
[1] "https://developer.github.com/v3/gists/#edit-a-gist"
```

What language is `nil`? I guess that's Ruby on their server.

gordonwoodhull:
Note that running `rcloud.update.notebook` directly works - afaict it's the same call?

```R
id <- rcloud.session.notebook()$content$id
rcloud.update.notebook(id, list(files=list('scratch.R'=NULL)))
```

gordonwoodhull:
If anyone needs a workaround, a colleague implemented it this way:

```r
dashboarddelete = function(name) {
        l = list()
        l[name]=list(NULL)
        rcloud.update.notebook(
            rcloud.session.notebook()$content$id,
            list(files=l))
    }
```

gordonwoodhull:
@bashlee, could you have a look at this one? Part of the effort is to determine how the current solution was supposed to work and see if it can be fixed. 

If not, the above workaround could be cleaned up and used.

PR #2544, opened 2018-01-24
fix for rcloud.delete.asset

bashlee:
Fixes att/rcloud#2420
Minor change in list passed to rcloud.update.notebook when deleting asset. Looks like the previous list was only handling the content part of the JSON rather than the file, ive added a patch work around.

### DIFF
--- a/rcloud.support/R/assets.R
+++ b/rcloud.support/R/assets.R
@@ -54,7 +54,7 @@ rcloud.execute.asset <- function(name, ..., notebook=rcloud.session.notebook(), 
   system(paste(driver, shQuote(asset), pars), TRUE, wait=wait)
 }
 
-rcloud.upload.asset <- function(name, content, notebook=rcloud.session.notebook(), binary=is.raw(content), file) {
+rcloud.upload.asset <- function(name, content, notebook=rcloud.session.notebook(), binary=is.raw(content), file, del = FALSE) {
     if (!missing(content) && !missing(file)) stop("content and file are mutually exclusive")
     if (!missing(file)) {
         file <- path.expand(file)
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Nice! It's unfortunate that we need this special case, but you're right that [the GitHub API](https://developer.github.com/v3/gists/#edit-a-gist) has this inconsistency.

gordonwoodhull:
Merged for 1.9.1. Thanks @bashlee!

s-u:
Wait - this is wrong way to do this! This is changing the original API - the inconsistency is the wrapping piece in the update part not in the functionality (c.f. the original delete.asset code!). The original contract is for NULL to be the deletion - this is "fixing" it at the wrong end.

gordonwoodhull:
The problem is that GitHub wants `"filename: null` not `"filename": {"content": null}`.

We could have the convention that when the content is null that triggers deletion by setting the whole object to null, and not need the `del` parameter, but the change still needs to be in `rcloud.upload.asset`

gordonwoodhull:
IMO it would be clearer what is going on if it read

```r
   if(del)
      l <- list(NULL)
   names(l) <- name 
```

instead of

```r
   if(del)
      l <- list(content=content)
   names(l) <- name
```

s-u:
Exactly - that's my point - it's a bug ion`upload.asset` where `NULL` gets wrapped incorrectly. But instead of fixing the bug this PR creates a whole parallel API that's not necessary and in the wrong place.

gordonwoodhull:
So you think that instead of having an extra argument `del`, `rcloud.upload.asset` should handle `content = NULL`?

I don't see a "whole parallel API that's not necessary and in the wrong place".

s-u:
Yes, that's what the official API is - only there was a bug in the wrapping of the deletion.
Parallel API is `del=...` since it's jsut another way to do the same and wrong place is `upload.asset` since that's what `delete.asset` is for.

gordonwoodhull:
So you think that `rcloud.upload.asset` should not handle any deletion cases, and `rcloud.delete.asset` should call `rcloud.update.notebook` instead of `rcloud.upload.asset`?

Issue #2519, opened 2017-11-29
updating a notebook doesn't change its position in the date-sorted tree


gordonwoodhull:
If the change is a rename, then the notebook moves into the correct position, as expected.

But if it's just changing a cell, adding a cell, etc, the notebook stays where it is even though its date has been updated:

![image](https://user-images.githubusercontent.com/1366709/33400365-e6534562-d523-11e7-832c-1624272b4775.png)

gordonwoodhull:
Similarly, if a notebook is renamed so that it enters an existing folder, the folder should be moved to the correct spot, not kept in the same spot as currently:

![image](https://user-images.githubusercontent.com/1366709/33400645-c9e3538a-d524-11e7-8f25-1b5ef7d99d17.png)

gordonwoodhull:
Finally, if the most recent notebook in a folder is renamed so that it leaves the folder, the folder should be moved to the spot determined by the next most recent notebook.

Issue #2500, opened 2017-11-18
show dates on folders when ordering by date


gordonwoodhull:
Related to #2497 defining folder date recursively, when the notebooks are ordered by date, the calculated date for each folder should be shown.

Perhaps (optionally?) also when filtering by date.

gordonwoodhull:
Fixed in 2.0

Issue #2520, opened 2017-11-29
updating a notebook that doesn't match date filter doesn't show the notebook


gordonwoodhull:
I know we discussed whether to show the current notebook in the tree if it doesn't match the date filter. There are good arguments either way on that one.

However, if modifying a notebook causes it to match the date filter, it definitely should be shown.

gordonwoodhull:
It appears that the current behavior is:
- if the filter is set while the current notebook does not match, that notebook is special-cased and shown
- if the filter is set and then an old notebook that does not match is loaded, the notebook is not shown. 
This is easy to repro using the Recent Notebooks menu.

This bug is still open.

PR #2545, opened 2018-01-25
Order by date tree fixes

shaneporter:
This PR addresses some date sorted tree issues.

#2500 - show dates on folders when ordering by date. I found my tree looked a little cluttered when folders have dates, so I also added a setting.
#2519 - notebook update causes update in date sorted tree.
#2497 - related to #2500, folders' dates are defined based on all their descendants


gordonwoodhull:
Good work! I think this folder ordering is very natural, and it's nice to see the folders moving to the top when a notebook inside is changed.

I agree, the display has more breathing room without the dates showing on folders, so it's good you put that under an option. I also notice the dates don't disappear again when changing back to alphabetical ordering. So maybe this is a misfeature - let's default it false.

Unfortunately, I noticed one bug which is making it hard to test a few of the cases in #2519. With this PR, I'm no longer able to move notebooks between folders by editing the name and changing the folder segment. For example, I changed the below notebook but it's not in its folder:

image

Also, if a notebook with a folder does not exist in the tree, and it's opened using Advanced / Load Notebook by ID, the folder fails to open and the notebook tree scrolls to the wrong position. You can test this by copying a notebook's ID and then deleting the notebook. Reload the page and then try loading the notebook with Load Notebook by ID.

(I was trying to test one of the previously issues, mentioned in https://github.com/att/rcloud/pull/2507#issuecomment-346152914, to see if old folders get put in the right place when they are added to the tree. It is in the right place! 😄  Just not too visible.)

shaneporter:
Thanks for reviewing @gordonwoodhull. I notice that just renaming the notebook (without moving folder) doesn't work in a 'name' sorted tree at all.

htdocs/js/tree/notebook_tree_view.js
```diff
@@ -281,21 +282,27 @@ RCloud.UI.notebook_tree_view = (function(model) {
...
-            var p = node.parent;
-            while(p.sort_order === this.model_.order.NOTEBOOK) {
-                this.$tree_.tree('openNode', p);
-                p = p.parent;
-            }
-            ui_utils.scroll_into_view(this.$tree_, 50, 100, null, $(node.element));
+            var that = this;
+            return new Promise(function(resolve) {
+                var p = node.parent;
+                while(p) {
```
shaneporter: Go up to the top, no?

shaneporter:
I'm hoping that things are a little more stable now. Renaming should now be ok, and the date sorted tree renames should jump around as appropriate.

I noticed a strange thing with the `scroll_into_view` - the `while` loop was checking only that the parent was a notebook, so when I had the tree completely collapsed, I never saw the relevant notebook. (See comment, above).

I also noticed that changing the tree sort type didn't 'find' the selected node either, so commit 18e3fec finds it again.

I started looking at #2520 and thought that it made most sense for the current notebook to be shown in the tree, regardless of filter. Whilst this contradicts the filter logic to some extent, I feel it's important to see the current notebook, since the user may wish to perform operations on it.  I made a change with 
559485d.

shaneporter:
The notebook dates should still be shown on the tree when it is sorted by name.

I'm not seeing the non-scrolling, 'out of sight' behaviour when a notebook's position is updated. On Chrome and Firefox, the notebook pane scrolls to the notebook's new position. 

I have made a code change so that the date sorted notebook position is updated when the notebook is modified.

gordonwoodhull:
Thanks @shaneporter, @dkey-pugh! 

I've reviewed this and merged to develop for 2.0

We'll continue to test and deal with any issues as they arise.

PR #2546, opened 2018-01-25
Update recent notebooks color coding rules

useless5771:
Couple of improvements to color-coding of recent notebooks list:
* removed duplicated functions which I must have missed out before
* made sure that removed notebooks are not included in the list of recent notebooks - they were displayed with 'null' user and 'something went wrong' description.
* added two more background colors styles

There are actually two implementations, you can switch between them by changing styleByLastAccessDate2 to styleByLastAccessDate in here: https://github.com/att/rcloud/compare/develop...MangoTheCat:features/issue-2462-2?expand=1#diff-2c2db8c6ac70968f344116281adbc075R546

styleByLastAccessDate performs color-coding by grouping the notebooks using fixed list of rules: 1,2,3,7,14,31 days from accessing the most recent notebook on the list
styleByLastAccessDate2 (default) performs color-coding by detecting 'gaps' between subsequent notebooks in the history. This implementation captures 'sequences' of notebooks accessed during a single 'session'. The classification rule is that two subsequent notebooks are in the same session if the gap between their access times is less than a given threshold. The threshold is 'dynamic' and for the first five groups is 8h, after that it is increased to 24h - this way first five groups represent 'days' and next groups represent weeks (and more).

I included both so you can test them and compare the results in an environment with some 'proper' activity history - my rcloud instance has too few notebooks with too large gaps between access times.

### DIFF
--- a/htdocs/editor_tab.js
+++ b/htdocs/editor_tab.js
@@ -426,7 +426,7 @@ var editor = function () {
         create_recent_notebooks_color_coder: function() {
           var backgroundColorStyler = function(i, element, dateWt) {
               var styles = [];
-              for(var j = 0; j < 5; j++) {
+              for(var j = 0; j < 7; j++) {
                 styles.push('recent-notebooks-group-' + j)
               }
               var component = 2;
@@ -434,7 +434,7 @@ var editor = function () {
               styles.forEach(function(style, i) {
                 $elem.removeClass(style);
               });
-              $elem.addClass(styles[~~((dateWt) * styles.length)]);
+              $elem.addClass(styles[~~((1-dateWt) * (styles.length-1))]);
             };
             
           if(this.color_recent_notebooks_by_modification_date_) {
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Sorry, but I don't understand the new algorithm. When I look at my own notebooks, all I see is a few notebooks green, yellow, orange, red, and then all the rest are purple out to infinity.

The algorithm we discussed is much simpler: just alternate between a few different colors based on the day. If the colors just end, that's not all that helpful.

image

Although quantizing to date isn't all that sophisticated, I think it captures the most common case, where people work during the day. So each day will appear as its own color, and there will continue to be daily sessions going all the way back in history.

I apologize this is taking a lot of iteration to get right. Clearly we didn't know what we wanted when we asked for this feature!

gordonwoodhull:
With `styleByLastAccessDate` I get two shades of green and then all yellow. Maybe you could produce some fake data to test this better?

image

useless5771:
@gordonwoodhull thank you for you review, indeed there was a bug in styleByLastAccessDate2 which was causing such odd result. Apologies for this.

I got some mock data injected into my instance now so I am able to test it with various use cases (thank you for sharing your notebooks history with me).

Fixed (not committed) styleByLastAccessDate2 produces the following result, it also cycles through colors now - ( I fixed it when I was working on inject mock data):

rcloud-recent-notebooks

The algorithm simply checks the gaps between notebook access times in the list, and if they differ in more than 8h it puts them to different 'buckets'. This of course doesn't work if someone works early in the morning and then in the evening - in such case the morning and evening 'sessions' will be colored differently - or if someone works late at night and starts early in the morning - in such case notebooks will have the same background.

I don't intend to commit it (unless you want to give it a go), as it does not implement the algorithm that we spoke about. I will work on it next.

Thanks

gordonwoodhull:
I see. If we just use an 8h gap all the way back, that should work well. Probably better than quantizing to days as I suggested.

Please update the PR and I'll try it out!

gordonwoodhull:
This works great, definitely well enough to be very useful. Thanks @useless5771! Rebased and merged for 1.9.1.

I added popup titles to the links to show the opened date and I think that will eliminate any confusion about what the coloring is about.

PR #2548, opened 2018-01-29
change code field to text_general

### LINKED ISSUE #2547: Indexing notebooks with long cells fails
The error from solr says that the code field is too long. Checking in the schema this is setup as a `string` field but I'm not sure why. The code field is a copyField that is not heavily used.

Fix inbound to change it to `text_general`.

dougmet:
Fixes #2547 

The code field is a copyField from content. Right now I'm not sure we're using it for much, so maybe it should be removed. I'm mainly leaving it in in case we want to add a custom filter for it.

Afraid I think it'll require a rebuild. Although might be able to use the [schema API](https://lucene.apache.org/solr/guide/6_6/schema-api.html#SchemaAPI-ReplaceaField) to run a quick test

### DIFF
--- a/conf/solr/schema.xml
+++ b/conf/solr/schema.xml
@@ -34,7 +34,7 @@
 <field name="raw_url" type="text_general" indexed="true" stored="true"/>
 <field name="size" type="int" indexed="false" stored="true"/>
 <field name="content" type="text_general" indexed="true" stored="true" multiValued="true"/>
-<field name="code" type="string" indexed="true" stored="true" multiValued="false" />    
+<field name="code" type="text_general" indexed="true" stored="true" multiValued="false" />    
 </fields>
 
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Thanks @dougmet! Indeed this fixes the cell length error.

This is merged for 1.9.1

gordonwoodhull:
Actually, we don't currently know how to restart solr without rebuilding it from scratch - we lose the solr core if we just do `solr stop` and `solr start` and we aren't sure if we want what they call "persistence" or "core discovery". 

It's something we want to ask you about tomorrow if you're available for the regular call.

gordonwoodhull:
I opened att/rcloud.solr#86 for the restart issue.

dougmet:
Hi, apologies @gordonwoodhull for silence. I'm looking into this now.

Issue #388, opened 2014-03-18
keep assets in alphabetical order


gordonwoodhull:
they jump around when reloading.

bashlee:
@gordonwoodhull  I can't seem to replicate this ? When adding a new asset and refreshing the session, all assets appear alphabetized

gordonwoodhull:
Right, the idea is to also add a new asset in the correct position, so that the order is always consistent and doesn't change on reload.

It isn't a big deal, but it's disorienting to have the order change on reload.

I think it should follow that rename also moves the asset to the correct position, since we implement rename as add + remove.

I don't think step 1 is necessary because we get the assets in alphabetical order from github / gist-services but please test that this is correct.

bashlee:
Makes sense, thanks for clarifying !

gordonwoodhull:
Fixed by #2550

PR #2550, opened 2018-01-31
Bug fix/ordering assets

bashlee:
New and renamed assets now appear in order without the need for page refresh att/rcloud#388

### DIFF
--- a/htdocs/js/notebook/notebook_model.js
+++ b/htdocs/js/notebook/notebook_model.js
@@ -35,14 +35,25 @@ Notebook.create_model = function()
                 return asset.filename() == filename;
             });
         },
-        append_asset: function(asset_model, filename, skip_event) {
+        append_asset: function(asset_model, filename, skip_event, user_appended) {
             asset_model.parent_model = this;
             var changes = [];
             changes.push(asset_model.change_object());
-            this.assets.push(asset_model);
+            var asset_filenames = _.map(this.assets, function(asset) { return asset.filename(); });
+            var new_asset_index;
+           
+            if(user_appended){
+                asset_filenames.push(asset_model.change_object().filename);
+                asset_filenames.sort();
+                new_asset_index =  asset_filenames.indexOf(asset_model.change_object().filename);
+                this.assets.splice(new_asset_index, 0, asset_model);
+                }           
+            else{
+                this.assets.push(asset_model);
+            }
             if(!skip_event)
                 _.each(this.views, function(view) {
-                    view.asset_appended(asset_model);
+                    view.asset_appended(asset_model, new_asset_index);
                 });
             return changes;
         },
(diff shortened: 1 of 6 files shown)


shaneporter:
https://github.com/att/rcloud/blob/9e3fa748a0e2f7fd1225a6326335e31f96564502/htdocs/js/upload_utils.js#L47

Adding a `true` parameter to the `append_asset` should sort it :)

bashlee:
@shaneporter  good spot, thanks ! I've pushed the change.

I've also noticed an issue with the ordering of capitalized assets. When both uploading and re-naming assets name with a capital the ordering does not appear logical - ill take a look at this before assigning back to Dave.

bashlee:
Cannot seem to reproduce issue I was seeing with capitalized assets, may be worth checking on test. 
Ready for test @dkey-pugh

shaneporter:
If the assets are default sorted, they will be ascii sorted, which will do this:

Apple, Axis, apple, banana

whereas appending `.sort((a,b) => { return a.localeCompare(b, undefined, { sensistivity: base; }); }` 

should give you the order you need. [More info on localeCompare can be found here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).

htdocs/js/notebook/notebook_model.js
```diff
@@ -35,14 +35,25 @@ Notebook.create_model = function()
...
             changes.push(asset_model.change_object());
-            this.assets.push(asset_model);
+            var asset_filenames = _.map(this.assets, function(asset) { return asset.filename(); });
+            var new_asset_index;
+           
+            if(user_appended){
+                asset_filenames.push(asset_model.change_object().filename);
+                asset_filenames.sort((a,b) => a.localeCompare(b, undefined, {sensitivity: 'base'}));
+                new_asset_index =  asset_filenames.indexOf(asset_model.change_object().filename);
+                this.assets.splice(new_asset_index, 0, asset_model);
```
gordonwoodhull: This wouldn't work unless they were already sorted (because both arrays have to have the same order), so I don't think the sort here is actually doing anything.

Since the assets are already in alphabetical order, it shouldn't be necessary to create a secondary array of all the filenames. 

You should be able to directly find the index of the first asset whose name is greater than the filename of the asset to be inserted, say using [Array.findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and the same `localeCompare` operation.

gordonwoodhull: I feel silly making you return to this after neglecting this for so long, so I'll make the above change when I test & merge tomorrow.

gordonwoodhull:
Thanks @bashlee! Merged for 2.0

gordonwoodhull:
Turns out GitHub sorts case-sensitive, so that was a false lead. 

It also looks like binary asset handling screws up the order even further, so first we get upper case, then lower case, then any binary assets. So we can't assume it's originally sorted at all.

I'll go ahead and fix these too.

gordonwoodhull:
Since the assets were in an untrusted order, I simplified things and always insert in (case insensitive) sorted order. It means we do an insertion sort when the notebook is loaded but that's the least of our concerns performance-wise.

PR #2554, opened 2018-02-05
Dataframe viewer improvements

shaneporter:
I'm not supposing that this is 100% complete, but a review would be good if you can spare some time @gordonwoodhull.

I've removed the padding from the viewer, styled the buttons like others in RCloud, and fixed the header. Horizontal scrollbars can't be eradicated when a dataframe has lots of columns.

I was considering a dialog. If the user wishes to look at the dataframe in more detail, they could click a 'magnifying glass' icon and see the dataframe full screen.

### DIFF
--- a/htdocs/js/ui/settings_frame.js
+++ b/htdocs/js/ui/settings_frame.js
@@ -231,21 +231,6 @@ RCloud.UI.settings_frame = (function() {
                             .filter(function(x, i, a) { return i==a.length-1 || !Notebook.empty_for_github(x); })
                             .join('/');
                     }
-                }),
-                'dataframe-page-size': that.text_input({
-                    sort: 13000,
-                    label: "Dataframe page size",
-                    default_value: "20",
-                    parse: function(val) {
-                        var defaultValue = 20, 
-                            parsed = parseInt(val);
-
-                        if(isNaN(parsed)) {
-                            return defaultValue;
-                        } else {
-                            return parsed < 1 ? 1 : parsed;
-                        }
-                    }
                 })
             });
         },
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
I agree the dialog may be helpful, but let's see how much we can do with just the session pane.

The styling is much nicer this way, and I especially like the way the headers stick but scroll top-level frame scrolls enough to let go of the title. This is just about perfect when there is enough horizontal space to display everything.

The pagination buttons scrolling into view still feels awkward to me, but I don't think we'd want those to be sticky unless they were really small, so let's ignore that for now.

I think the big thing missing right now is that it's not grabbing all the vertical space it can. Panels like Assets, Help, and Notebooks have a flag set somewhere (is it `panel_sizer`?) so that they take up the available space. If the other panel are minimized, Dataframe should not need vertical scrolling.

There's also a bunch of extra whitespace below the pagination buttons.

gordonwoodhull:
Per @s-u: 
 * Layout should more compact so that more data fits on the screen
 * Pagination size persisted as a user option (in particular, the ability to see 100 rows instead of the 20 we have now).

gordonwoodhull:
The page size control (with options 10, 20, 50, 100, 200 I guess) should be inside the dataframe panel, but the setting should be saved as a regular user setting. Just like we do with the notebook tree view settings. Is this possible?

I haven't looked at DT in depth, but it looks like they have a page size control; we'd just have to detect what it's set to.

gordonwoodhull:
Also, I'm not seeing page size take, it just always shows 20 rows.

shaneporter:
Quite why I thought of not using the DT paging is still a mystery to me.

gordonwoodhull:
Thanks @shaneporter, this is looking really great!

I think @s-u will want this even more compact. Is it possible to change the layout so that the page size chooser fits at the bottom alongside the buttons but above or below the "information"? (which is great to have!)

shaneporter:
`padding: 0` on the table cells now, so I've added some borders to the `th` and `first-child` `td`s for some visual distinction (things can get pretty cramped at lower widths).

The 'information' now sits under the page length dropdown. The paging sits on the right side.

gordonwoodhull:
Thanks @dkey-pugh, @shaneporter! 

I've reviewed this and merged it for 2.0.

PR #2557, opened 2018-02-13
Move the solr home to solrdata. att/rcloud.solr#86

dougmet:
This relates to att/rcloud.solr#86

Main changes are:

* add `solr.xml` so that the `solrdata` directory can be used as the solr home
* update `conf/solr/solrsetup.sh` with the new path locations

Before we merge, I just wanted to check what the expected behaviour for `$SOLR_HOME` should be? I don't have that set on my machine so I need to start solr like this:

```sh
/data/rcloud/solr/solr/bin/solr start -s /data/rcloud/solr/solrdata
```

And this will survive restarts. In `solrsetup.sh` I check for `$SOLR_HOME` and will use that path if the variable has been set.

### DIFF
--- a/conf/solr/solr.xml
+++ b/conf/solr/solr.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+ contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


s-u:
looks good!

gordonwoodhull:
Thanks @dougmet, @s-u! Rebased and merged for RCloud 1.9.1

Issue #2551, opened 2018-02-05
autoscroll can't keep up if output is too fast


gordonwoodhull:
If the output is lots of individual lines without too much delay between them, the output will get ahead of the autoscroll's ability to track it.

gordonwoodhull:
I imagine this will be browser-specific, but on my mac running Firefox Developer, it seems to be able to keep up when the delay is 0.25s but not 0.2s, with the following test:

```r
for(i in 1:1000) {
    print(1:1)
    Sys.sleep(0.25);
} 
```

gordonwoodhull:
I'm able to repro on Chrome but I have to set the delay to 0.1 instead of 0.2s.

gordonwoodhull:
Fixed by #2559 in 2.0

PR #2559, opened 2018-02-23
Improve notebook output autoscroll functionality

### LINKED ISSUE #2522: Indexing a notebook fails
with 

```
Error in rjson::fromJSON(comments_json) : 
  STRING_ELT() can only be applied to a 'character vector', not a 'list'
trace:
.Call("fromJSON", json_str, unexpected.escape, PACKAGE = "rjson")
rjson::fromJSON(comments_json)
build_comments(notebook_info)
build_update_metadata(notebook, starcount)
rcloud.solr::update_solr(rcloud.get.notebook(id), star.count)
```

useless5771:
Introduce results buffering and process the results in batches. This removes fleakyness which was caused by race conditions between: R producing the results and pushing htmlwidgets contents, autoscroll feature and user scroll actions.

Thanks to this only scrolls up the notebook cell area need to be detected and considered as a trigger for disabling autoscroll.

The jquery.scrollto library got updated to alternative implementation which is actively maintained, accepts more parameters and exposes an option to stop the scrolling animation on user scroll.

FIX #2522, #2551

This change needs to be applied together with: https://github.com/att/rcloud.htmlwidgets/pull/24

Tested with various use cases, like:
```
for(i in 1:300) {
    print(1:1)
    Sys.sleep(0.001);
} 
```
```
require(DT)
for(i in 1:100) {
    print(1:1)
    if((i %% 10) == 0) {
        stocks <- data.frame(
              time = as.Date('2009-01-01') + 0:104,
              X = rnorm(105, 0, 1),
              Y = rnorm(105, 0, 2),
              Z = rnorm(105, 0, 4))
            
        print(datatable(stocks))
        
    }
    Sys.sleep(0.1);
} 
```

I tested the deferred results using the following script
```
for(i in 1:10) {
    print(1:1)
    print(wplot(1:10, 1:10))
} 
```
this was the only function that I found to provoke this type of results, I guess this is mainly used when computation separation is turned on? @gordonwoodhull if you can provide me with some examples that  use deferred results and that I could run in my dev environment I am happy to test the changes in this PR against them.

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -30,8 +30,17 @@ function create_cell_html_view(language, cell_model) {
     var highlights_;
     var code_preprocessors_ = []; // will be an extension point, someday
     var running_state_;  // running state
-    var running_ui_state_ = {add_result: {}};
     var autoscroll_notebook_output_;
+    var results_processing_context_ = {
+      options : {
+        no_of_results_in_batch: 20,
+        notebook_update_delay : 20,
+      },
+      results : [],
+      stop : false,
+      prev_scroll : null,
+      scrolled : false
+    };
 
     // input1
     var prompt_text_;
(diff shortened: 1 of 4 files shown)


gordonwoodhull:
Thanks @dkey-pugh, @useless5771!

I've reviewed and merged for 2.0

Issue #1334, opened 2015-02-16
provide run time inputs in python code


sanketd11:
In python cell execution, if any of the statement in a code requires run time user inputs(e.g. raw_input()'), the execution remains stuck at that statement. The stop button remains enabled until the page reload

gordonwoodhull:
@sisukapalli looked into this but it didn't make it into this release.

gordonwoodhull:
@useless5771, as part of your Jupyter work, please look into how they do `input()`, `raw_input()`, etc. and see if you can hook it into the same UI handler we use for R stdin, e.g. `readline()`

gordonwoodhull:
Implemented in RCloud 2.0

Issue #2565, opened 2018-03-21
multiple languages in a single extension package


gordonwoodhull:
Currently each language extension (`rcloud.r`, `rcloud.python`, ...) implements exactly one language.

Rather than having to create a dummy package for each language supported by Jupyter, why don't we allow each package to return multiple languages when `rcloud.language.support()` is called?

<strike>I think the idiomatic R way of doing this is to return vectors for each field (`language`, `run.cell`, `ace.mode`, etc.) instead of a single value. [setup.R](https://github.com/att/rcloud/blob/e46ab0a4db66779f60390b43ce153a5db08c31b7/rcloud.support/R/setup.R#L355) will have to do more complicated validation/unpacking of the data, but it can be backward compatible. ("Scalar" fields will have the same type, since scalars are length-1 vectors in R; function fields will require a type test, since a vector of function references is different from a function reference.)</strike>

What's really nice about this is that `rcloud.jupyter` can detect at startup time which languages are available, and return that set dynamically.

gordonwoodhull:
Per @s-u, the entry point should return either one (S3) object or a list of them. The class tag is mostly to identify if you got a single object or a list back.

dkey-pugh:
@gordonwoodhull what's the scope on testing this? Execution, std in read, code completion and syntax highlighting for python2, python3 and julia? Java/Javascript? R/Scala/Python Spark - can we test this with a 'dummy' kernal or will we need an Apache Spark instance? There's a lot of functionality being added in one issue and I'd like to be sure we're on the same wavelength over testing it.

gordonwoodhull:
Thanks @dkey-pugh. Yes, that is about the scope of it - those four features across 7-8 languages.

This is *the* major feature in RCloud 2.0 so yes, the fact that it's one issue is deceptive.

A dummy kernel for Spark will be fine, as long as you can test the language features. It's not an exercise in cluster management, although you'll probably get some more practice in system administration!

PR #2566, opened 2018-03-26
Switch from IPython to Jupyter

### LINKED ISSUE #1378: upgrade ipython2 to jupyter
We are using a modified `notebook_runner.py` which worked with IPython 2.

The idea is to upgrade this to the current one so that it can work with IPython 3 a.k.a. Jupyter.

I don't know how difficult this is but I'll start tracking links here.

If we had this we could wrap rcloud.python's `rcloud.language.support` function in a generator which takes a language kernel name, and then generate language supports instead of only using it for python.

useless5771:
Jupyter integration is implemented using nbconvert API and its ExecutePreprocessor class which implementation was customized to support code completion, standard input prompts and reuse of kernel instances.

For details on configuration see README.MD file.

FIX #1378, #1334


gordonwoodhull:
Wow, fantastic!

Per #2565, I think we should rename the package, since it should be able to support many languages soon. I don't know why anyone would want to use the old `rcloud.python` package, but I think we should leave it intact and create a new package.

`rcloud.jupyter`, I suppose, even though it's somewhat ambiguous.

Couple little comments on the source. Hope to test this with Dave's help over the next week.

rcloud.packages/rcloud.python/R/main.R
```diff
@@ -1,59 +1,112 @@
-.start.python <- function(rcloud.session)
+#' 
+#' Configuration
+#' 
+#' Python version and library path are defined by the following properties:
+#'  * rcloud.python.path - path to Python installation that should be used
+#'  * python.extra.libs - additional Python lib directories (e.g. where Jupyter Python modules got installed)
+#'
```
gordonwoodhull: Please mention these are `rcloud.conf` settings. (Not that we have many other kinds of properties.)

rcloud.packages/rcloud.python/R/main.R
```diff
@@ -1,59 +1,112 @@
...
+  }
+  
+  inline_plots <- "%matplotlib inline"
+  paste0(retina, '\n', inline_plots)
+}
+
+#'
+#' Initializes Jupyter Adapter and stores reference to it in session with 'jupyter.adapter' key.
+#'
+.start.jupter.adapter <- function(rcloud.session)
```
gordonwoodhull: please fix typo in function name

rcloud.packages/rcloud.python/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Python 3
+
+```{bash}
+sudo python3 -m pip3 install jupyter
+```
+
+### Configuring RCloud Jupyter integration
+
+RCloud exposes the following properties:
```
gordonwoodhull: again, maybe "uses these settings from `rcloud.conf`"

useless5771:
Thank you @gordonwoodhull  for you review, I addressed the typos and improved the documentation.

Regarding renaming the package/creating a new one. I thought of doing this as part of #2565 but of course since rcloud.python should remain I will do this now. 

Would you like the rcloud.jupyter package to be put into separate git repository as well?

Thanks

gordonwoodhull:
I think it's fine to keep it here, but if you prefer to move it that's an option too. It's a fuzzy line "what is core RCloud?"

useless5771:
I think it would make sense to put it to separate repository, but I would leave it for now where it is to avoid cross-repository commits as part of #2565. We can move it to separate repository after that.

useless5771:
I am closing this PR, it will be superseded by a PR that adds new rcloud.jupyter package instead of editing existing one.

PR #2567, opened 2018-03-27
Introduce rcloud.jupyter package

useless5771:
Main differences to what was included in PR https://github.com/att/rcloud/pull/2566 are:

* renamed rcloud.conf properties to rcloud.jupyter.*
* renamed some of the functions so they mention 'jupyter' instead of 'python' (where applicable)
* Improved README.md and included steps covering registration of rcloud.jupyter as rcloud.language


gordonwoodhull:
Subsumed by #2570

PR #2570, opened 2018-04-24
Expose Jupyter kernels as RCloud languages 

useless5771:
Jupyter integration is implemented using nbconvert API and its ExecutePreprocessor class which implementation was customised to support code completion, standard input prompts and reuse of kernel instances.

* Introduced RCloudLanguage class representing language support (before simple list was used)
* Exposed Jupyter kernels as languages.
* Exposed configuration of Jupyter kernel mapping to RCloud languages.
* Introduced Jupyter_completer ACE editor component which collects completions produced by Jupyter kernels.
* some minor tweaks/improvements in the way ACE editor Mode instances are created and configured 
* fixed code completions handling produced by vanilla ACE code completers

For details on configuration and available ACE editor modes see REAME.MD file.

This change also fixes worker-(css|html|javascript).js which update was omitted when ACE editor was upgraded to 1.2.6 in RCloud.

FIX #2565

PR #2576, opened 2018-06-01
Fix invalid method name.

useless5771:
In Jupyter adapter script one of the method was named the same way as objects attribute, because of this kernels were never stopped.

Relates to #2570


useless5771:
@gordonwoodhull this is the fix to jupyter kernel shutdown issue.

gordonwoodhull:
Thanks @useless5771! Merged for 2.0

PR #2586, opened 2018-07-10
Pass through content-disposition header.

useless5771:
Exposed rcloud.proxy.headers.passthrough property defaulting to 'Content-Disposition' which can be used to specify a list of headers that should be passed through by proxy.html.

FIX https://github.com/att/rcloud.shiny/issues/28


gordonwoodhull:
Merged for 2.0.1. - thanks @useless5771!

PR #2587, opened 2018-07-10
Enable result consumer on std in read.

### LINKED ISSUE #2585: jupyter input is hanging
We implemented input for Python in #1334, and I'm pretty sure it worked, but now it hangs (for both Python2 and Python3 cells).

I'm not sure, but it looks like it may have to do with results buffering for autoscroll. At least, the UI goes into an infinite promise loop in `stop_results_consumer`.

IIRC the two features were developed on parallel branches. So they probably work separately. (Disabling autoscroll is not enough to fix it, however.)

useless5771:
The issue was that the result consumer was not started by the code handling the standard input request

FIX #2585

### DIFF
--- a/htdocs/js/notebook/cell_view.js
+++ b/htdocs/js/notebook/cell_view.js
@@ -39,6 +39,7 @@ function create_cell_html_view(language, cell_model) {
       results : [],
       stop : false,
       prev_scroll : null,
+      result_consumer : null,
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
I'm going to merge this to develop without testing, since I have no idea how to get Jupyter running on any of my testing machines!

PR #2591, opened 2018-08-21
Improve handling of errors during notebook creation.

useless5771:
Small improvement to handling errors that may happen during creation of a notebook, before when creation of a notebook failed, the error was masked by error thrown at line 379:

```
ulog("INFO: rcloud.create.notebook (", res$content$id, ")")
```

I also added a page covering configuration of gist service so it supports forking/import of notebooks larger than 10MB: https://github.com/att/rcloud-gist-services/wiki/LargeNotebooksSupport


gordonwoodhull:
Thanks @useless5771!! Merged for 2.0.1

PR #2592, opened 2018-08-31
Features/issue merge e127

### LINKED ISSUE #127: notebook/gist merging
I'm not sure if this is another thing that would require digging down to the git API, but I'd like to consider the possibility of selectively merging changes from other notebooks.

In particular, I'm looking at the use case where dev A has created a function/notebook F and documented it in the README.md cell we're proposing. But not very well. Now smart articulate user B likes the function but thinks the documentation stinks. So B forks notebook F and fixes the README. 

Now how does A take the doc changes? And what if the implementation of F has changed in the meantime? Afaik there are no pull requests for gists. Can we do better than copy and paste here?

useless5771:
Notebook Merge functionality from e/127 branch aligned with current state of the develop branch.

FIX #127


gordonwoodhull:
This feature has a huge surface area, so we will definitely need to iterate on it. I'm going to point out the biggest problems now. Once we've solved those, let's merge to develop and continue to iterate.

For the most part, this works surprisingly well for such a complicated feature - thanks @shaneporter and @useless5771 for all your hard work! 

Here's are the bugs I'm seeing (on my Linux test machine):
* The editor view is sized wrong on both Chrome and Firefox and is pretty much not visible. On Firefox it's very short 

image

and on Chrome it is very skinny

image

* When merging from a notebook that has more cells, the interface properly shows that cells will be added. However, going the other way, when the source notebook has less cells, no action is indicated, and none happens. Probably it should indicate that the cells will be deleted (and should do so). 

* One time when performing multiple merges in a single session, it appeared that the notebook state was cached. The merge indicated that no changes would be made even though the source notebook had changed and was now different. However, I have not been able to repro this one yet.

gordonwoodhull:
A couple of improvements I'd like to see to the interface:
* If there are no changes at all, the interface should indicate that nothing will happen. Preferably it shouldn't commit a change either, but this is less important.
* As Mats suggested, a way to choose All or No changes from a file. Currently the default is All, and you have to opt out of each change. But it would be better to be able to say "don't change this file" - and then once you can do that, you also need a way to go back and say "actually yes do select all the changes."

I'm also not finding the icons and colors completely clear. I would expect green to mean additions and red to mean deletions, but in the file list it means Take or Don't Take. Also the rewind icon is clever but pretty weird. :) I will have to think about this and look at other tools to make some suggestions. A merge interface is always a confusing and complicated thing, so I want to tread carefully here. 

It's amazing that the interface is as clean and clear as it is, and I don't to mess it up by adding a lot of other buttons and options.

useless5771:
@gordonwoodhull thank you for your comments, I will fix the issues that you've found so far and update this PR.

useless5771:
The latest bulk of commits should address most of the issues. 

The new icons look like this:
merge-feature

merge-feature-2

And when there are no changes to merge the 'Merge' button is disabled:
merge-feature-3

Outstanding issues are:
* dialog layout in Firefox and Chrome.
* cached state (I experienced this as well, but don't have a a consistent way of reproducing this (yet)...)

gordonwoodhull:
Thanks @useless5771, this is looking really good. The new icons are much more clear.

I still think the interface is going to be difficult for a lot of people, but this is just fundamentally difficult stuff. I am only glad we didn't try to do 3-way merging in the first round!

I was able to test on Chrome / MacOS today, where there are no layout issues. It works very well. Firefox 62 & 63 on MacOS still have height ~5px, but Safari 12 also works.

I have a slew of minor changes I'd like to make, but let's start with a few big interface changes:
 - Both Merge Changes dialogs have a default action indicated with the blue button, so <kbd>enter</kbd> should execute it.
 - Let's rename the second Merge Changes dialog "Apply Changes", at least for now. That might not be the best title, but definitely the two dialogs should have different titles so that it's possible to talk about them.
 - I don't think the initial view is helpful in the "Apply Changes" dialog. The animation is cute, but really, why stretch that much content across the whole window? The Editor/Diff area should be showing from the start, even if there are no changes.
- When there are no changes, we should indicate that more prominently. Disabling the Merge button is good (arguably it change to "Nothing to do" instead) but I think we also need to say why. (A column of equals signs will bewilder new users). So how about displaying ***no changes*** in bold grey italic in the editor/diff area?
- @s-u also requested that all the changes be shown at once. Instead of the summary table choosing which cell to show changes for, it would become more of a navigation tool, scrolling through all the changes on the right. This may be too difficult! but please see if this is at all feasible. 

I'm resisting digging into things like "should we display 1/2 when we're taking one of out two change?s", and "should the action text in the Editor show what will currently happen, or what clicking will change the action to?", "do we really need the Diff pane?", but be forewarned, that kind of stuff is coming next... UI design can be pretty tedious.

I think if we address the above then we will be ready to get feedback.

useless5771:
The dialog now displays diffs of all files in a scrollable pane:

merge-feature-4

In case of large files, their diff editors will be expanded up to the scrolable area viewport height:

merge-feature-5


When a file is excluded from the merge, the corresponding diff panel becomes transparent:

merge-feature-6



Changes made to the UI also include:
* colors are a bit different - they are aligned with the panel styles from bootstrap.
* removed 'theirs' column - it seemed redundant as in most cases it contained the same value as 'yours'.
* Removed header from the file list table


What is outstanding:
* more testing in Chrome, 
* I just did very limited testing in FF, 
* linting
* I don't like how accepting/rejecting specific changes in a diff editor provokes tooltips in all editors to refresh, will need to investigate this.

gordonwoodhull:
This works great! In a brief test of a few merges between a few notebooks, I did not see any problems at all.

I will go ahead and merge this to develop for 2.1, and start filing issues for any UI tweaks. (Hopefully someone else can take that up!)

PR #2603, opened 2018-10-15
Turn off detailed diff generation for large files.

useless5771:
In case of larger files (greater than 0.25MB) generating a diff resulted in a long execution of JS and/or crashing a web browser.

If a file is greater than 0.25MB the diff representing a replacement of whole file contents is generated to avoid the above issues.

This change also improves the performance of generating diffs for new and deleted cells/assets as the diff library isn't unnecessarily executed for such files.

Some minor linting of API.

@gordonwoodhull this is a set of changes that I hadn't finished off before https://github.com/att/rcloud/pull/2592 got merged


gordonwoodhull:
Thanks @useless5771!

Rebased and merged for 2.1

PR #2604, opened 2018-10-18
Changes required by rcloud.params package

useless5771:
Add JS result type

This type of result represents a JS function executed by cell result processing loop.


gordonwoodhull:
I find the name for this result type misleading. 

To me it sounds like it will append JavaScript code to the cell, which is not what happens - it's really delaying a function call until any previous content is output.

What do you think about calling it function_call or something like that?

useless5771:
Thank you @gordonwoodhull , you are right, the result type indeed was misleading. I renamed it to the 'function_call'.

I also included another commit that fixes an issue with displaying forms that use bootstrap classes - it was impacting styling of rcloud.params input controls.

gordonwoodhull:
Thanks @useless5771! Looks good.

PR #2607, opened 2018-10-25
Bugfix/Issue 2598

JABedford:
When renaming a pre-existing folder in the tree, an error log would display as there was a scoping issue with the update_notebook_from_gist function.


gordonwoodhull:
This doesn't look right. Did you test it?

`rcloud.update_notebook` is an OCAP, an R function tunneled into JS. It takes a gist id, the notebook changes, and a boolean as parameters.

Additionally, `update_notebook_from_gist` isn't a global, so I think this will throw an undefined variable error.

I think the underlying problem is that the function on `editor` object has been renamed or moved, so the solution should be at the call which is failing rather than the call before that.

PR #2610, opened 2018-10-26
Bugfix/Issue-2577

JABedford:
The this.$tree_ is undefined error was eventually found to be being triggered by selecting the same notebook from a different directory. 

Assigning "this" to a new variable fixes the issue.

### DIFF
--- a/htdocs/js/tree/notebook_tree_view.js
+++ b/htdocs/js/tree/notebook_tree_view.js
@@ -283,9 +283,10 @@ RCloud.UI.notebook_tree_view = (function(model) {
         },
 
         select_node: function(node) {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Makes sense. That looks like it would break every time, wonder why it ever worked.

Thanks @JABedford!

gordonwoodhull:
One little trick you can do that will help me out, is to include `fixes #xxx` or `closes #yyy` in your commit messages when they solve an issue. GitHub [will automatically close](https://help.github.com/articles/closing-issues-using-keywords/) issues when it sees certain messages in commits that are merged to the default branch.

PR #2611, opened 2018-11-01
Bugfix/issue 2608

### LINKED ISSUE #2608: eliminate "inner scrolling" in Apply Changes dialog
If there is a big change, the editor doesn't go above a certain height and starts scrolling. It isn't possible to see all the changes just by scrolling the window; one has to also be careful to scroll the content within each change.

Two level scrolling like this is always confusing to users.

Each Monaco editor should be tall enough to show all its content, the same way we do in the RCloud notebook interface using Ace.

JABedford:
Fixes #2608 

Fix for the issue of inner scrolling in the editor panel by allowing the editor to grow depending on the amount of changes.

### DIFF
--- a/htdocs/js/tree/notebook_tree_controller.js
+++ b/htdocs/js/tree/notebook_tree_controller.js
@@ -59,8 +59,8 @@ RCloud.UI.notebook_tree_controller = (function(model, view) {
         },
 
         has_notebook_info: function(gistname) {
-            return this.model_.has_notebook_info(gistname);
-o        },
+            return this.model_.has_notebook_info(gistname);        
+        },
 
         get_notebook_info: function(gistname) {
             return this.model_.get_notebook_info(gistname);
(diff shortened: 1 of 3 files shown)


gordonwoodhull:
Hmm, it looks like now the sections grow to the height of the containing div, but not necessarily far enough to show all of the changes.

For example, I'm replacing a long cell with another long cell here:

image

There are about 10 lines of the content to be deleted which are missing here, as well as another section which will be added.

I'm not sure, but I think that setting the height at 100% will use the height of the containing element, whereas we need to use the height of the *content*. 

Actually, if I just remove `max-height` from `.diff-panel` then it seems to show everything - does this make sense?

JABedford:
Strange. I can only assume for some reason maxHeight is being assigned as height in your circumstance. These are the variables that control the content. That wasn't my intention and it probably would have been the wiser option to remove it completely.

Please find changes below Gordon and advise if working correctly for yourself.

JABedford:
Max-height removed from the SASS file Gordon.

Irrelevant now, but funnily enough the previous commit worked fine in Firefox but not in Chrome. Lesson learned!

gordonwoodhull:
Aha! I should have thought of that when we were seeing totally different things. I can confirm that the original commit works on Firefox but not Chrome. 

Firefox is more standards-compliant, but I'm in the habit of testing on Chrome first because the developer tools are still so much better. (Even though I use Firefox as my main browser.)

Funny that the major browsers don't completely agree on the basic CSS Box model even in 2018. I think that shows how complicated this stuff is.

Rebased and merged for 2.1. Thanks @JABedford!

Issue #2613, opened 2018-11-01
Line numbers in margin of Apply Changes dialog don't make sense


gordonwoodhull:
The line numbers just continue from one added/deleted section to the next, which doesn't reflect either source notebook or the eventual result.

<img width="247" alt="image" src="https://user-images.githubusercontent.com/1366709/47858920-766cc080-ddc3-11e8-80f7-3b756ff950a4.png">

Either the line numbers of the source notebooks should be displayed here, or we should not display line numbers at all.

PR #2618, opened 2018-11-12
Remove unwanted Monaco decorations

### LINKED ISSUE #2616: tiny scrollbar and diff annotations are visible and interactive in Apply Changes
It's difficult to dig into this since it seems right-clicking is disabled somehow. Monaco's scrollbar and some annotations are visible in the right gutter of the diffs:

image

When the page first loads I am able to drag the tiny scrollbar. The only effect is that the cursor moves and some grey boxes appear. I presume these are diff annotations, but they don't seem to align with what we are doing.

This is not as urgent a problem as some of the other things we are looking at, but it still would be nice to completely hide the scrollbar now that #2608 is fixed.

JABedford:
This hides the various Monaco decorations by moving the main editor panel across the unwanted parts. Much cleaner!

Fixes #2616 and #2613


gordonwoodhull:
Thanks @JABedford! Looks great!

PR #2621, opened 2018-11-19
Display Notebook ID in Info Popup and allow user to Copy to Clipboard.

### LINKED ISSUE #2594: add notebook id to notebook info popup
It's useful to be able to get at a notebook id without opening it (e.g. for merging etc).

Mats pointed out that an obvious place to put this is the notebook info popup that you get when clicking the (i) icon in the tree.

Let's put it there, and find a simple cross-browser way to add a Copy button next to it. (You used to have to use flash for this but I think there are legit ways to do it now.)

JABedford:
Notebook ID now displays in the pop up information and can be copied to the users clipboard ready to use in the merge dialogue.

Fixes #2594

### DIFF
--- a/htdocs/sass/components/rcloud-base.scss
+++ b/htdocs/sass/components/rcloud-base.scss
@@ -1180,3 +1180,49 @@ body .modal-dialog.full {
         }
     }
 }
+
+/* Styles specific to the Copy-To-Clipboard functionality 
+    for the Notebook Info panel. */
+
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Awesome! It works great and will be very helpful.

We might consider compacting the popup as it's taking up a lot of vertical space, but let's go with this for now.

Merging for 2.1. Thanks @JABedford!

PR #2622, opened 2018-11-20
Merge Into Current Notebook Button & Folder Delete Fix

### LINKED ISSUE #2597: merge into current notebook button
As suggested by @s-u: even easier than getting the notebook id with #2594, why not add a button to each notebook in the tree that starts the merge process with that notebook?

We have been using the "branch" icon to mean "fork" because it looks similar, and Font Awesome doesn't have the real "fork" icon e.g. seen above in the GitHub interface. It doesn't look like a true merge icon is available in the free version of Font Awesome, but we might be able to flip the branch icon vertically.

### LINKED ISSUE #2612: Incorrect behaviour of "Remove Folder" dialogue.
When a user tries to remove a folder, and then cancels the dialogue, it is remembered if they then try at any point to remove the same folder.

screenshot 2018-11-01 at 14 12 20

JABedford:
Fixes #2597 & Fixes #2612 

The merge button is displayed in the tree, and when clicked on will open a new merge dialogue. The ID of the notebook that the user clicked on will automatically be present in the ID input box allowing the user to continue.

I wasn't sure whether the intention of this feature was to take the user straight through to the "Stage Changes" dialogue, but actually I think directing to the initial merge screen is perhaps a more graceful solution as it gives the user a "confirmation" screen, rather than staging the changes, incase the Merge button was clicked on accidentally.

I added the fix to the folder renaming issue just because it was a quick fix in the same file I had been working in.

### DIFF
--- a/htdocs/js/ui/notebook_commands.js
+++ b/htdocs/js/ui/notebook_commands.js
@@ -1,5 +1,6 @@
 RCloud.UI.notebook_commands = (function() {
     var icon_style_ = {'line-height': '90%'};
+    var merge_icon_style_ = {'line-height': '90%', 'padding-right': '3px'};
     var star_style_ = _.extend({'font-size': '80%'}, icon_style_);
     var star_states_ = {true: {'class': 'icon-star', title: 'unstar'},
                         false: {'class': 'icon-star-empty', title: 'star'}};
@@ -205,6 +206,9 @@ RCloud.UI.notebook_commands = (function() {
         },
         icon_style: function() {
             return icon_style_;
+        },
+        merge_icon_style: function() {
+            return merge_icon_style_;
         },
         decorate: function($li, node, right) {
             var appeared;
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
I think it's fine to go straight to Apply Changes, since there is still no action taken at that point. There isn't anything interesting the user could do at the Merge Changes dialog, and they can still cancel by clicking the button or pressing <kbd>esc</kbd>.

I think that people may use the Merge Notebook feature as a way to quickly bring up a diff, so all the better to go straight to it.

(The "current merge source" for the notebook should still be updated when the button is clicked.)

Speaking of accidental clicks, the Merge button is very close to the Remove button. I guess this has to do with the shape of the icon; it wasn't an issue when hide/show notebook was in that spot. It looks like the framework for adding buttons has the ability to customize the style, so could you please add a tiny bit of padding there?

Finally, I think "merge" is somewhat ambiguous - could you please change the popup to read "merge here"? (Or something... "merge into this notebook" would be too long.)

Overall this looks great and will be very useful. I'll wait on these tweaks to merge it.

Thanks @JABedford!

gordonwoodhull:
Thanks @JABedford! Looks and works great!

Merged for 2.1

PR #2623, opened 2018-11-20
Accidental Globals in editor_tab.js.

### LINKED ISSUE #2590: accidental globals from editor_tab refactor
I noticed that `username_` and `current_` are not defined in `editor_tab.js`, so JavaScript has promoted them to global variables.

Don't have time/patience at the moment to find the right tool to lint all our code for this. (Thought jshint would do this, but it's curiously silent about this problem.)

JABedford:
Fixes #2590 

Is incorporating a tool something we are looking at doing immediately @gordonwoodhull ? ESLint seems to be the most popular choice currently. Perhaps we could create a new dedicated issue to discuss further?


gordonwoodhull:
Thanks @JABedford.

Ah the whims and fads of Javascript... Yes ESLint sounds good.

Sure, I'll open another issue.

gordonwoodhull:
Thanks @JABedford, merged for 2.1

PR #2627, opened 2018-11-28
Stop cell results consumer if there are no results in the queue. 

useless5771:
Remove code responsible for waiting for delayed results and replace it with function_call result.

> Just the last commit is relevant to this PR, the rest is covered by att/rcloud/pull/2614

This change is complementary to: att/rcloud.htmlwidgets/pull/26 and removes previous workaround for the  race condition.

Unfortunately the changes didn't decrease CPU use, it is still at ~40% for blocking rcloud.params form - some further investigation is required. I did some further ad-hoc profiling by turning off animations and background htmltools resizing callbacks, but it didn't make much difference nor revealed obvious suspects.


gordonwoodhull:
Thanks Mats! I am going ahead and merging the changes to core, params, and htmlwidgets.

There is further cleanup that Simon and I will have to do but we can take it from here.

Thanks for your conscientious work on all of this.

PR #2637, opened 2018-12-06
Merge Notebook from File Tree triggers ReferenceError.

### LINKED ISSUE #2634: Merge Notebook from File tree triggers ReferenceError?
When usig the Merge Here feature from the file tree instead of the Advanced menu, when the diff is displayed, the cell contents hang and the console shows "Uncaught ReferenceError: monaco is not defined at init".
This blocks merger via this workflow.

JABedford:
The workflow from the tree bypasses the original Monaco require statement, therefore Monaco is undefined.

Not sure why it worked so well without it!!


gordonwoodhull:
Thanks @JABedford, it looks like the first Merge Changes dialog was loading Monaco, so probably speeding through that dialog caused a race condition.

gordonwoodhull:
Replaced with same change at a higher scope. Thanks @JABedford!

PR #2638, opened 2018-12-06
'Invisible' Editor for file undefined error during Notebook merger.

### LINKED ISSUE #2635: 'Invisible' Editor for file undefined error during Notebook merger
When a user tries to merge two notebooks using the Advanced menu, on generating the diff, the console displays an error. This is not reflected in the UI and does not appear to affect the merger (i.e. the user is unaware that it happens if the console is not open).

The error is "Editor for file undefined and model $model<x> not found!" where <x> is an integer that appears to increment by two in each session each time the error is displayed.

JABedford:
Fixes #2635 

This is being caused by a console.error statement. Everything seems to work fine even if the error is thrown so I have just commented it out for now, and perhaps warrants some further investigation?

### DIFF
--- a/htdocs/js/ui/notebook_merger/merger_view.js
+++ b/htdocs/js/ui/notebook_merger/merger_view.js
@@ -551,7 +551,7 @@ RCloudNotebookMerger.view = (function(model) {
             })[0];
             let editor_descriptor = this._editors[model_filename];
             if(!editor_descriptor) {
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
The question is, why is it reporting an error?

gordonwoodhull:
Looked into this a bit more, still think there are some extra editors hiding somewhere unused.

I'll take this fix for now. Thanks @JABedford!

PR #2640, opened 2018-12-10
Not all codelens show up when going straight to Apply Changes

### LINKED ISSUE #2639: not all codelens show up when going straight to Apply Changes
Once we load the Monaco library on time #2634, seems like we still miss the first few codelens. There is blank space allocated for them but they are not there.

image

Second time loading Apply Changes, they are all there.

JABedford:
Fixes #2639 

As mentioned in the issue, reinserting the require statement seems to solve this error.

Some light linting included as well.

### DIFF
--- a/htdocs/js/ui/notebook_merger/merger_view.js
+++ b/htdocs/js/ui/notebook_merger/merger_view.js
@@ -69,6 +69,13 @@ RCloudNotebookMerger.view = (function(model) {
       //
       //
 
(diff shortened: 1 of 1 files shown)


gordonwoodhull:
Sorry, but this doesn't fix the issue for me. I should have mentioned that I removed the extra require because it is redundant and incorrect, as noted in #2639.

I think there may be some confusion because this problem only occurs the very first time you try a merge, using the button in the notebook tree.

PR #2720, opened 2020-03-23
sanitize errors in notebook.R to prevent reflected XSS

gordonwoodhull:
Hi @s-u, requesting your review on this fix we discussed last week.

I cleaned it up slightly by sanitizing the output rather than the inputs. Although notebook.R can `return()` other errors, I don't think they are vulnerable to XSS. I haven't reviewed the other scripts, but I would propose fixing them the same way.

I bet there is a better way to do multiple replacements but I didn't find it.

### DIFF
--- a/htdocs/notebook.R
+++ b/htdocs/notebook.R
@@ -104,6 +104,6 @@ run <- function(url, query, body, headers)
       list(payload, type)
     }
   }, error=function(e) {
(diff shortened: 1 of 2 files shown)


gordonwoodhull:
Silence is consensus, or something. Merging for 2.2.3

PR #2734, opened 2020-05-21
rcloud.conf option github.use.query.token

gordonwoodhull:
Hi @s-u, here is the RCloud part of att/rgithub#3.

I have tested
- old RCloud works fine with new rgithub and rcloud-gist-services
- this change works fine with github.com as the backend and does not include the `access_token` query parameter it was complaining about

I have not tested this change with rcloud-gist-services.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,11 @@
  * Sharable Link tooltip shows the view type (#2723)
  * New Cell buttons are faster (#2731)
 
+### Features
+ * Compatible with HTTP Basic Authentication for `github.com`. The option is `github.use.basic.auth`
+   and defaults `false` for compatibility with
+   [rcloud-gist-services](https://github.com/att/rcloud-gist-services)
+
 ### Bugfixes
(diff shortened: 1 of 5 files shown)


gordonwoodhull:
Based on feedback from @s-u, I changed the name from `github.use.basic.auth` to `github.use.query.token`, with the opposite meaning and opposite default; see also att/rgithub#4.

It also defaults the flag based on `rational.githubgist`.

Currently this requires one of those two to be set, since `github.use.query.token` defaults true like the rgithub setting.

However, on further consideration, we could default it false at the RCloud level (even though defaulted true in rgithub for backward compatibility). Then rcloud.conf would be by default compatible with GitHub, although rgithub can’t be.

gordonwoodhull:
@s-u, I think this and att/rgithub#4 satisfy your concerns.

If you have time to review, please give them a look.

gordonwoodhull:
I'm merging this because I think it satisfies the concerns raised (and I got the deprecation warning again).

We can always revisit before release.

PR #2770, opened 2025-04-15
include zig deps in fat tarball, add example and instructions

pomi601:
To work around the Zig 0.14 package manager bug which prevents it from successfully fetching dependent Zig packages when run behind an HTTP proxy, I've added code to include the zig cache directory into the fat binary distribution, and instructions and a sample script.


s-u (approved):

s-u:
Great, thanks!

PR #2773, opened 2025-04-29
download zig release version 0.14.0 by default

pomi601:
Updates download.sh script to download Zig release version 0.14.0 if no argument is given.


s-u:
Thanks!

PR #2774, opened 2025-05-03
readme: add section on bypassing fetch-assets

pomi601:
Adds a section on how to bypass `fetch-assets` if there's a network or repository problem. The straight answer is quite simple (although it comes with serious security implications), but I wanted to make sure to explain all the background, because there are some subtleties developers might need to understand. Ideally, there would be separate documentation of the build system as its own entity, and the README could be more succinct, referring to relevant sections for additional background. I'm open to suggestions on how to improve this.


s-u (approved):

s-u:
Makes sense, thanks!

PR #2777, opened 2025-06-26
workaround: Docker Compose stage config: ensure rcloud.home() directory is user-writeable

pomi601:
Without this workaround, the `stage` Docker Compose configuration does not provide a user-writeable package directory in `.libPaths()`, which makes package installation to the archivable data directory impossible.

I suspect there is some other configuration error during the rcloud startup that isn't setting directory permissions as I expect, but I wasn't able to track that down.

The other change in the diff is a docker build cache optimization, to move the declaration of ARG RCLOUD_APT_INSTALL a bit lower in the file, so earlier apt install stages don't need to be repeated when changing the value of the argument.

### DIFF
--- a/docker/common/Dockerfile
+++ b/docker/common/Dockerfile
@@ -243,11 +243,6 @@ RUN --mount=type=cache,target=/data/rcloud/zig/cache,sharing=locked  \
 #
 FROM base AS runtime
 
-# Additional system requirements to install when building the image.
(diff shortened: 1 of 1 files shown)


s-u (approved):
Looks good - the fact that the home directory permissions are not set looks like a bug in that particular code path, since in the authenticated case that is taken care of.

